PR #2, opened 2015-03-17
Fix for using the out argument from the console

mfunkie:
Updating reference to argv.o to be the output variable we set.
Updating output to point to argv.out instead of argv.o because abbreviations aren't the representation on the argv object


fkling:
Thansk! Release as v1.0.2. I also added test cases for the CLI in 68526e8e938b1fab77fa832109809acd4bc0e32d and found another issue through that.

Issue #6, opened 2015-03-18
Support ES6 export declarations


fkling:
Make the default resolvers understand `export` declarations, i.e.

```js
import React from 'react';
export default React.createClass(...);
// or
export var Component = React.createClass(...);
// etc
```

banderson:
@fkling Is this something that's already underway? I can lend a hand if needed, this would be huge for me.

fkling:
I haven't started yet. If you want to take a stab on it, I'd appreciate it!

banderson:
Ok, great, I'll see what I can do.

For those who need this ASAP and want a clue of where to start, I spiked out the minimal functionality I needed [here](https://github.com/banderson/react-docgen/commit/d21f8f2e65b83ca7ef708fc6f139f14a36585054). This just covers allowing ES6 style imports and _default_ exports. Hopefully I can eventually flesh out the remaining details and get it merged in here.

fkling:
@banderson: Did you make any progress with this? What's the status?

banderson:
@fkling I got `React.createClass` recognition for ES6 `default exports` and ID'ing `React` in ES6 `import` statements in just that small change linked above. This works well for my use cases. Now I'm working on `PropTypes` recognition for using named imports (e.g. `import React, {PropTypes} from 'React';`) and in restructured assignment (`let {PropTypes} = React;`).

Before this is done, we'd need to ensure all import export cases are covered, and tests for all the existing cases :/ Glad to help out as much as I can, would welcome help from anyone else that knows more use cases if they exist :)

fkling:
Great, thanks! Feel free to make a PR for the things that are working. The destructuring assignment case should already be covered with 25399598ed3d5ebbdb13ed3707183751f257ccd3.

fkling:
This should work as of 623c1923a89329ede0cad138333fc3c481158492, release with v1.3.0

PR #12, opened 2015-04-30
For #6 Support ES6 export declarations

morlay:
#6 

```` js
import React from 'react';
export default React.createClass(...);
// or
export var Component = React.createClass(...);
// etc
````` 
This way is done.

Should we support this?
```` js
import React, { createClass } from 'react';
export default createClass(...);
````` 

And not sure, is it ok, to support ES6 module and CommonJS in one handler.

### CODE REVIEW

**Comment by fkling** [comment]
Sweet! I will take a close look later today. Thank you for this!

File: src/resolver/findExportedReactCreateClassCall.js [hunk 1]
```diff
@@ -47,6 +48,17 @@ function findExportedReactCreateClass(
     visitDoWhileStatement: ignore,
     visitForStatement: ignore,
     visitForInStatement: ignore,
+    visitExportDeclaration: function(path) {
+      path = resolveToCallExpression(path);
```
fkling: Mmh, do we have to introduce `resolveToCallExpression`? My approach would have been to just extract the path from `ExportDeclaration` that we are interested in, and run `resolveToValue` on that.

fkling:
It doesn't seem like this would support support

```js
export {
  comeponent,
  something else
};
```

or does it?

> Should we support this?
>
    import React, { createClass } from 'react';
    export default createClass(...);

Not sure. I think if the need arises we can still that.

> And not sure, is it ok, to support ES6 module and CommonJS in one handler.

Seems fine to me.

src/resolver/findExportedReactCreateClassCall.js
morlay: Sure, seems we can change here with 
``` js
path = resolveToValue(path.get('declaration'));
```
and add the conditions to `resolveToValue`
``` js
if (types.VariableDeclaration.check(node)) {
    return resolveToValue(path.get('declarations', 0))
  } else if (types.VariableDeclarator.check(node)) {
     if (node.init) {
       return resolveToValue(path.get('init'));
     }
  } 
```

I will push another commit to update the change. 

however, I feel the `resolveToValue` need to do some refact now, when we add a lot of conditions.

morlay:
Cool. I feel we are not necessary to write like that.

With react-docgen's help, we can get docs so easy, in my last projects. 

So I put the docs tool in my [testing repo](https://github.com/morlay/react-semantic-ui) ([preview](http://morlay.github.io/react-semantic-ui/components-doc))

Recently, we need to update to ES6, that why I try to make the update for react-docgen.

fkling:
Sorry for the late response! I included your changes in 55c1511cdd192f51f7183ed7ddc604ba7e86a68e and 9e8670c7b8197d7be98af8bf1d2b21d2763fced3 (I squashed the third one).

I also built on top of your changes and added support for the remaining `import` and `export` variations.

All these changes are release with v1.3.0.

Thanks for your help!

PR #28, opened 2015-10-06
Stateless Components

iamdustan:
Starts support for Stateless Component detection and documentation. Currently this adds an `isStatelessComponent` util method. I’ll start working on working this in to the resolvers now, but wanted early review on this to see if I missed anything.

### DIFF
--- a/src/__tests__/main-test.js
+++ b/src/__tests__/main-test.js
@@ -99,4 +99,81 @@ describe('main', () => {
     `);
   });
 
+  describe('Stateless Component definition: ArrowFunctionExpression', () => {
+    test(`
+      import React, {PropTypes} from "React";
+
+      /**
+        * Example component description
+        */
+      let Component = props => <div />;
+      Component.displayName = 'ABC';
+      Component.defaultProps = {
+          foo: true
+      };
+
+      Component.propTypes = {
+        /**
+        * Example prop description
+        */
+        foo: PropTypes.bool
+      };
+
+      export default Component;
+    `);
+  });
+
+  describe('Stateless Component definition: FunctionDeclaration', () => {
+    test(`
+      import React, {PropTypes} from "React";
+
+      /**
+      * Example component description
+      */
+      function Component (props) {
+        return <div />;
+      }
+
+      Component.displayName = 'ABC';
+      Component.defaultProps = {
+          foo: true
+      };
+
+      Component.propTypes = {
+        /**
+        * Example prop description
+        */
+        foo: PropTypes.bool
+      };
+
+      export default Component;
+    `);
+  });
+
+  describe('Stateless Component definition: FunctionExpression', () => {
+    test(`
+      import React, {PropTypes} from "React";
+
+      /**
+      * Example component description
+      */
+      let Component = function(props) {
+        return React.createElement('div', null);
+      }
+
+      Component.displayName = 'ABC';
+      Component.defaultProps = {
+          foo: true
+      };
+
+      Component.propTypes = {
+        /**
+        * Example prop description
+        */
+        foo: PropTypes.bool
+      };
+
+      export default Component;
+    `);
+  });
 });
--- a/src/__tests__/main-test.js
+++ b/src/__tests__/main-test.js
@@ -13,10 +13,11 @@
 jest.autoMockOff();
 
 describe('main', () => {
-  var docgen;
+  var docgen, ERROR_MISSING_DEFINITION;
 
   beforeEach(() => {
     docgen = require('../main');
+    ({ERROR_MISSING_DEFINITION} = require('../parse'));
   });
 
   function test(source) {
@@ -175,5 +176,38 @@ describe('main', () => {
 
       export default Component;
     `);
+  });
+
+  describe('Stateless Component definition', () => {
+    it('is not so greedy', () => {
+      const source = `
+        import React, {PropTypes} from "React";
+
+        /**
+        * Example component description
+        */
+        let NotAComponent = function(props) {
+          let HiddenComponent = () => React.createElement('div', null);
+
+          return 7;
+        }
+
+        NotAComponent.displayName = 'ABC';
+        NotAComponent.defaultProps = {
+            foo: true
+        };
+
+        NotAComponent.propTypes = {
+          /**
+          * Example prop description
+          */
+          foo: PropTypes.bool
+        };
+
+        export default NotAComponent;
+      `;
+
+      expect(() => docgen.parse(source)).toThrow(ERROR_MISSING_DEFINITION);
+    });
   });
 });
--- a/src/resolver/__tests__/findAllComponentDefinitions-test.js
+++ b/src/resolver/__tests__/findAllComponentDefinitions-test.js
@@ -168,11 +168,17 @@ describe('findAllComponentDefinitions', () => {
           var result = () => <div>{props.children}</div>;
           return result();
         };
+        const ComponentF = function(props) {
+          var helpers = {
+            comp() { return <div>{props.children}</div>; }
+          };
+          return helpers.comp();
+        };
       `;
 
       var result = parse(source);
       expect(Array.isArray(result)).toBe(true);
-      expect(result.length).toBe(6);
+      expect(result.length).toBe(7);
     });
 
     it('finds React.createElement, independent of the var name', () => {
--- a/src/resolver/__tests__/findAllComponentDefinitions-test.js
+++ b/src/resolver/__tests__/findAllComponentDefinitions-test.js
@@ -160,11 +160,15 @@ describe('findAllComponentDefinitions', () => {
         var Obj = {
           component() { if (true) { return <div />; } }
         };
+        const ComponentD = function(props) {
+          var result = <div>{props.children}</div>;
+          return result;
+        };
       `;
 
       var result = parse(source);
       expect(Array.isArray(result)).toBe(true);
-      expect(result.length).toBe(4);
+      expect(result.length).toBe(5);
     });
 
     it('finds React.createElement, independent of the var name', () => {
--- a/src/resolver/__tests__/findAllComponentDefinitions-test.js
+++ b/src/resolver/__tests__/findAllComponentDefinitions-test.js
@@ -149,4 +149,46 @@ describe('findAllComponentDefinitions', () => {
     });
   });
 
+  describe('stateless components', () => {
+
+    it('finds stateless components', () => {
+      var source = `
+        import React from 'React';
+        let ComponentA = () => <div />;
+        function ComponentB () { return React.createElement('div', null); }
+        const ComponentC = function(props) { return <div>{props.children}</div>; };
+        var Obj = {
+          component() { if (true) { return <div />; } }
+        };
+      `;
+
+      var result = parse(source);
+      expect(Array.isArray(result)).toBe(true);
+      expect(result.length).toBe(4);
+    });
+
+    it('finds React.createElement, independent of the var name', () => {
+      var source = `
+        import AlphaBetters from 'react';
+        function ComponentA () { return AlphaBetters.createElement('div', null); }
+        function ComponentB () { return 7; }
+      `;
+
+      var result = parse(source);
+      expect(Array.isArray(result)).toBe(true);
+      expect(result.length).toBe(1);
+    });
+
+    it('does not process X.createClass of other modules', () => {
+      var source = `
+        import R from 'FakeReact';
+        const ComponentA = () => R.createElement('div', null);
+      `;
+
+      var result = parse(source);
+      expect(Array.isArray(result)).toBe(true);
+      expect(result.length).toBe(0);
+    });
+  });
+
 });
--- a/src/resolver/__tests__/findAllComponentDefinitions-test.js
+++ b/src/resolver/__tests__/findAllComponentDefinitions-test.js
@@ -164,11 +164,15 @@ describe('findAllComponentDefinitions', () => {
           var result = <div>{props.children}</div>;
           return result;
         };
+        const ComponentE = function(props) {
+          var result = () => <div>{props.children}</div>;
+          return result();
+        };
       `;
 
       var result = parse(source);
       expect(Array.isArray(result)).toBe(true);
-      expect(result.length).toBe(5);
+      expect(result.length).toBe(6);
     });
 
     it('finds React.createElement, independent of the var name', () => {
--- a/src/resolver/findAllComponentDefinitions.js
+++ b/src/resolver/findAllComponentDefinitions.js
@@ -12,6 +12,7 @@
 
 import isReactComponentClass from '../utils/isReactComponentClass';
 import isReactCreateClassCall from '../utils/isReactCreateClassCall';
+import isStatelessComponent from '../utils/isStatelessComponent';
 import normalizeClassDefinition from '../utils/normalizeClassDefinition';
 import resolveToValue from '../utils/resolveToValue';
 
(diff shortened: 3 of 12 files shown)


src/utils/__tests__/isStatelessComponent-test.js
```diff
@@ -0,0 +1,130 @@
...
+    it('accepts simple function declaration components', () => {
+      var def = parse('function Foo () { return <div /> }').get('body', 0);
+      expect(isStatelessComponent(def)).toBe(true);
+    });
+  });
+
+  describe('Stateless Function Components with React.createElement', () => {
+    it('accepts simple arrow function components', () => {
+      var def = parse(
+        'var Foo = () => React.creatElement("div", null);'
```
fkling: Looks good so far. One remark is actually covered by a typo in this test :D (note the missing `e` in `React.creatElement`.

Seems like it accepts any function which calls another function as react component.

src/utils/isStatelessComponent.js
```diff
@@ -0,0 +1,76 @@
...
+
+function containsJSXElementOrReactCreateElementCall(node) {
+  var visited = false;
+  recast.visit(node, {
+    visitJSXElement(path) {
+      visited = true;
+      this.traverse(path);
+    },
+
+    visitCallExpression(path) {
```
fkling: See above, that would consider any function that calls another function as component. Is that intentional?

src/utils/isStatelessComponent.js
```diff
@@ -0,0 +1,76 @@
...
+  // 'ArrowExpression', 'ArrowFunction'
+  'ArrowFunctionExpression'
+];
+
+function containsJSXElementOrReactCreateElementCall(node) {
+  var visited = false;
+  recast.visit(node, {
+    visitJSXElement(path) {
+      visited = true;
+      this.traverse(path);
```
fkling: We wouldn't need to traverse deeper if we already found our answer. You could just `return false;` to indicate that the traversal is done for this node.

src/utils/__tests__/isStatelessComponent-test.js
iamdustan: ha, well, the issue is that I accidentally deleted the check to ensure that I was calling `createElement` in the visitor below so basically everything is matched :innocent: 

I’ve gotten a bit stuck on trying to duplicate `isReactCreateClassCall` to `isReactCreateElementCall` which handles ensuring the correct createElement method is called and it looks up to the root import React bit to ensure it’s the real react.

in `src/utils/resolveToValue.js` `path.scope.lookup(node.name)` is always returning undefined.

src/utils/isStatelessComponent.js
```diff
@@ -0,0 +1,76 @@
...
+  // https://github.com/estree/estree/issues/2
+  // 'ArrowExpression', 'ArrowFunction'
+  'ArrowFunctionExpression'
+];
+
+function containsJSXElementOrReactCreateElementCall(node) {
+  var visited = false;
+  recast.visit(node, {
+    visitJSXElement(path) {
+      visited = true;
```
fkling: Mmh, that would also consider `foo` in 

```javascript
function foo() {
  return function bar() {
    return <div />;
  }
}
```

as React component, i.e. any function that contains JSX is considered to be a React component. I wonder if we can narrow this down more.

src/utils/__tests__/isStatelessComponent-test.js
fkling: In your test? You'd also have to add `var React = require('react');` to the test code.

iamdustan: Yeah, I’ve done that to each...well, technically I’ve done `import React from 'react';` or `import R from 'react';`.

I’m bouncing for dinner, but should have the next push sometime tomorrow with the next set of questions :)

fkling: :+1: Really appreciate your effort :)

iamdustan: Oh! Found my sticking point. Since I’m calling `recast.visit(node, {})` in the l`containsJSXElementOrReactCreateElementCall` function I’m losing all of recasts internal state

iamdustan: Managed to get through this. Passing the root `path` instead of `path.node` didn’t lose state.

src/utils/isStatelessComponent.js
iamdustan: Yeah, I’ve thought of this as well. After I get through the current cases with detecting propTypes (last step I believe) then I’ll look into maybe doing something like a return value lookup of each function rather than the current heuristic of “contains JSX or React.createElement”. I suspect this could get a bit tricky or annoying or untenable if return value is something imported from another module, though. I intend on not supporting that :)

iamdustan:
I notice you have `class` components normalize to `static propTypes` rather than <ES6-style properties. I suspect I’ll need to the propTypeHandler to also read from <ES6-style properties for stateless components unless you have any other ideas.

```
function Component (props) {
  return <div />;
}

Component.propTypes = {};
```

fkling:
I attribute it to bad design decisions at the beginning ;) Everything kind of expects that the result of the resolver is a single AST node and extracts data from there.

I guess we could introduce some new interface, `ResolverResult`, and have implementations for `class`, object literals (`React.createClass`) and stateless functions, to abstract from the different representations. Could be quite some work though.

Handlers shouldn't necessarily what kind of definition they are dealing with. That custom logic could be implemented in the `getMemberValuePath` method though.

iamdustan:
I’m currently working on extending the `getMemberValuePath` to also `getMemberExpressionValuePath` and see how far that takes me. I think it’ll only need to handle variable declarations, function expressions, and function declarations. This should avoid needing an new interface and will continue building on the early design decisions :)

All of my API transformation projects have terrible design decisions. And be “terrible” I mean “no”.

fkling:
> All of my API transformation projects have terrible design decisions. And be “terrible” I mean “no”.

:)

iamdustan:
Well....I think this is actually working now.. :open_mouth:

iamdustan:
hmmm....not quite working, but close.

iamdustan:
Alright, fixed my final commit. This is now working for stateless function components in my current codebase as expected :)

iamdustan:
@fkling I probably have until EOD to finish this up then will have to bow out for a while. I want to prioritize supported use cases versus not supported use cases. This is the known issues ordered by what I think the priority should be. What do you think?

* `isStatelessComponent` is too loose in saying true. Should scope this better (high priority)
```
const NotAComponent = () => {
  const Component = () => React.createElement('div', null);
  const AlsoAComponent = () => React.createElement('div', null);

  return {Component, AlsoAComponent};
}
```
* hash of components (low/no priority)
```
const bag = {
  Component() { return React.createElement('div', null); },
  const AlsoAComponent = () => React.createElement('div', null);


```

iamdustan:
Also e4c867ce1524222734ef0a9daa4a01623f8caee4 is of note for the `integration`-style test proving that is indeed working correctly. :)

src/resolver/findAllComponentDefinitions.js
```diff
@@ -34,7 +35,19 @@ export default function findAllReactCreateClassCalls(
...
+      // TODO: normalizeStatelessDefinition to pick up propTypes
+      definitions.push(path);
+    }
+    return false;
+  }
+
   recast.visit(ast, {
+    visitFunctionDeclaration: statelessVisitor,
+    visitFunctionExpression: statelessVisitor,
+    visitProperty: statelessVisitor,
```
fkling: I don't think we need this, do we? What case does this capture which isn't captured otherwise?

fkling:
This is pretty awesome work, thank you so much! Regarding priorities:

> isStatelessComponent is too loose in saying true. Should scope this better (high priority)

Definitely that one. Not sure how tricky this one is though. 

> hash of components (low/no priority)

I wonder what the use cases are for doing something like this. This would be covered by finding all component definitions anyway though, wouldn't it?

---

I was wondering about this following case:

```javascript
function Foo() {
  return <div />;
}

Object.assign(Foo, {
  displayName: 'Foo',
  propTypes: { ... }
});
```

Do you think that would be a common use case? Or are `propTypes` and `displayName` the only properties one would add to stateless components? I guess `defaultProps` too. Maybe this is a "nice to have".

src/resolver/findAllComponentDefinitions.js
iamdustan: I added this as part of supporting some of the permutations also being added to gaearon/babel-plugin-react-transform; namely this format: https://github.com/gaearon/babel-plugin-react-transform/pull/34/files#diff-9d6f6ff3fa6b5b63c09190bd7a2efb95R15

Honestly I’m not sure if that is captured or handled correctly yet. :innocent:

fkling: I think this should be handled by the other visitors since methods are just represented as function expressions in the AST.

On Oct 7, 2015, at 12:51 PM, Dustan Kasten <notifications@github.com<mailto:notifications@github.com>> wrote:


In src/resolver/findAllComponentDefinitions.js<https://github.com/reactjs/react-docgen/pull/28#discussion_r41437429>:

>    recast.visit(ast, {
> +    visitFunctionDeclaration: statelessVisitor,
> +    visitFunctionExpression: statelessVisitor,
> +    visitProperty: statelessVisitor,


I added this as part of supporting some of the permutations also being added to gaearon/babel-plugin-react-transform; namely this format: https://github.com/gaearon/babel-plugin-react-transform/pull/34/files#diff-9d6f6ff3fa6b5b63c09190bd7a2efb95R15

Honestly I’m not sure if that is captured or handled correctly yet. [:innocent:]

iamdustan:
> isStatelessComponent is too loose in saying true. Should scope this better (high priority)

I have a couple ideas to try. The first is just to filter on all ReturnStatement’s in the body of a possible stateless component and detect if it’s a ReactElement. Though I’m not certain if these will be considered valid components off-hand. +@spicyj if I need to account for those. I don’t think they are at current..

```
const IsThisAComponent => null;
const IsThisAComponent => 'string';
const IsThisAComponent => 7;
const IsThisAComponent => [<div key="first" />, <div key="last" />];
```
The weakness is this will definitely fail for the following permutations:

```
const Component(props) {
  if (Math.random() > 0.5) return <div />;
  else return <span />;
}

const Component(props) {
  return (Math.random() > 0.5) ? <div /> : <span />;
}
```

Alternatively I could do some scope crawling and ensure that `JSXElement` and `React.createElement` are in not in a nested scope. This would handle the following case:

```
var HOC = (options) => {
  const Component = (props) => <div />
  Component.propTypes = {};
  return {Component};
}
```

Sadly, I think that to be totally accurate we’ll need to track all possible return values and ensure each is a `JSXElement` or `React.createElement`. E.g. lookup return values for both `truthyFunc` and `falsyFunc` in the following example:

```
const truthyFunc = () => <div />;
const falsyFunc = () => <span />;
const Component = (props) => props ? truthyFunc() : falsyFunc();
```

spicyj:
@iamdustan Those aren't currently valid components but will be (at least the one with `null`) in 0.15.

iamdustan:
thanks @spicyj. I’ll punt that until 0.15 :)

fkling:
In the worst case we can also introduce a doclet, like `@react` or `@react-component` to mark functions as React components that would be hard to identify otherwise.

iamdustan:
@fkling this now supports a pretty wide array of use cases. It’s not quite as bullet proof as I’d like, but this is all I have within me for now. Let me know what you think and if you want to block on anything.

src/resolver/__tests__/findAllComponentDefinitions-test.js
```diff
@@ -149,4 +149,60 @@ describe('findAllComponentDefinitions', () => {
...
+          var result = () => <div>{props.children}</div>;
+          return result();
+        };
+        const ComponentF = function(props) {
+          var helpers = {
+            comp() { return <div>{props.children}</div>; }
+          };
+          return helpers.comp();
+        };
+      `;
```
iamdustan: This function for this isn’t quite recursive enough. While all of these resolve as expected, the following two cases do not resolve correctly yet:

```
const ComponentG = function(props) {
  var nested = {
    helpers: {
      comp() { return <div>{props.children}</div>; }
    }
  };
  return nested.helpers.comp();
};
const ComponentF = function(props) {
  var other = () => <div>{props.children}</div>;
  var helpers = {
    comp: other
  };
  return helpers.comp();
};
```

src/__tests__/main-test.js
```diff
@@ -99,4 +100,114 @@ describe('main', () => {
...
+          */
+          foo: PropTypes.bool
+        };
+
+        export default NotAComponent;
+      `;
+
+      expect(() => docgen.parse(source)).toThrow(ERROR_MISSING_DEFINITION);
+    });
+  });
```
iamdustan: Here is my failing test case for the bag of Components. I’m not terribly worried about this right now, personally.

```
  xdescribe('Stateless Component definition: in an Object', () => {
    test(`
      import React, {PropTypes} from "React";
      /**
      * Example component description
      */
      let components = {
        Component(props) {
          return <div />;
        }
      };
      components.Component.displayName = 'ABC';
      components.Component.defaultProps = {
          foo: true
      };
      components.Component.propTypes = {
        /**
        * Example prop description
        */
        foo: PropTypes.bool
      };
      export default components.Component;
    `);
  });
```

iamdustan:
@fkling I’m good here unless you come up with additional code examples that fail. A thorough review to undo the cobwebs I’ve created over the past week could likely help clean up the code in here.

src/__tests__/main-test.js
iamdustan: the bag of components still fails. I’m personally okay with this for a first iteration and if folks open an issue about it then it can be tackled.

src/resolver/__tests__/findAllComponentDefinitions-test.js
iamdustan: This works like a champ.

fkling:
@iamdustan: I'll try to get to it tomorrow or Wednesday!

iamdustan:
Another thought I had while adding so many use cases was: do we even want to support all the *possible* ways to do stateless components for this? The idea of these stateless function components is for simplicity. Documenting only a subset of these could help encourage folks to keep things simple.

fkling:
> do we even want to support all the possible ways to do stateless components for this?

Yep, good point. My answer to this is: no. We probably already don't support every possible way to create a React component and that's OK IMO.
A side effect of this project is that people have to follow certain guidelines to make it work, which in turn hopefully positively impacts consistency / familiarity across different projects / code bases.

Not sure I can get to it this week (will be away tomorrow for the rest of the week) and there is another issue I'd like to address before releasing a new major release (and I think this should be a major release because it could pickup components that haven't been considered as such).

iamdustan:
:thumbsup: can’t slow me down! https://github.com/iamdustan/react-docgen/tree/stateless-components-dist :)

src/resolver/findAllComponentDefinitions.js
```diff
@@ -34,7 +35,18 @@ export default function findAllReactCreateClassCalls(
     return false;
   }
 
+  function statelessVisitor(path) {
+    if (isStatelessComponent(path)) {
+      // TODO: normalizeStatelessDefinition to pick up propTypes
```
fkling: This is done right? (because we use a different way to resolve this information)

iamdustan: Why yes I do believe it is!

jherr:
:+1:

iamdustan:
Friendly ping for review.

fkling:
Sorry, I thought I would be able to get something else in, but don't really have much time. I'm not so sure anymore this requires a major version bump. In the worst case it just finds more components.

A minor version bumpv should suffice an which case I can get this in now. What do you think?

iamdustan:
I’ve been using it without issues since the PR was opened so it feels generally safe to me. It shouldn’t cause any regressions on `createClass` or `extends React.Component` so it should be fine with just a minor, IMO.

fkling:
Merged in 9f27006d297f2184403e8bf5842dd64da99e632b and released as v2.4.

Thank you so much for this!

iamdustan:
Woohoo! Made it. Thanks!

PR #32, opened 2015-10-16
Enable es7.doExpressions and es7.functionBind

blai:
They are part of Stage 0 features: https://babeljs.io/docs/usage/experimental/#stage-0


fkling:
Last I heard is that function binding might not advance to the next stage in its current form, but there is no reason why react-docgen should have an opinion on this.

Thanks!

PR #46, opened 2015-12-11
Update main-test.js

oliviertassinari:
Test that the evaluation is working for the defaultProps and not only propTypes.

### DIFF
--- a/src/__tests__/main-test.js
+++ b/src/__tests__/main-test.js
@@ -151,6 +151,36 @@ describe('main', () => {
     `);
   });
 
+  describe('Stateless Component definition: FunctionDeclarationEvaluation', () => {
+    test(`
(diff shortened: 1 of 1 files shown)


fkling:
Given that the test fails, it seems that `defaultProps` is not resolved to its value?

oliviertassinari:
Exactly.

donnieberg:
@fkling Thanks for brining this to their attention. I'm using ES6 classes and am experiencing the same problem with getting defaultProps documentation from my components.

fkling:
I fixed the issue locally and simply changed the existing tests to use referenced propTypes and defaultProps, instead of adding your additional test.

I hope that's fine with you, thank you very much for noticing this issue!

PR #48, opened 2016-01-05
Support for flow types

### LINKED ISSUE #31: flow types resolver
```jsx
import React from 'react';
import type {UserProperties} from '../types/UserProperties';

type UserProps = {
  // the users full name
  name: string;
  // the users age
  age: number;
  // user properties like bio and location information
  properties: UserProperties;
};

/**
 * Render a User
 */
const User = (props:UserProps) => (
  <div>{props.name} {props.age} years old<UserProps {...props.properties} /></div>
);
```

would become something like:

```
{
  "description": "Render a User",
  "props": {
    "name": {
      "type": {
        "name": "string",
        "value": "string"
      },
      "required": true,
      "description": "the users full name",
      "defaultValue": {
        "value": ""
        "computed":false
      }
    },
    "age": {
      "type": {
        "name": "number",
        "value": "number"
      },
      "required": true,
      "description": "the users age",
      "defaultValue": {
        "value": ""
        "computed":false
      }
    },
    "properties": {
      "type": {
        "name": "UserProperties",
        "value": "UserProperties"
      },
      "required": true,
      "description": "user properties like bio and location information",
      "defaultValue": {
        "value": ""
        "computed":false
      }
    },
  }
}
```

danez:
I started working on adding support for flow types.

So far I only tried to get it working for classes, and it already works nicely for this usecase, but only for string type and declared types. The code although is far away from nice yet. It also needs a patched version of ast-types, which I will create a PR also soon.

I opened this PR already so that interested people can join and already have a look and maybe give feedback.

Will fix #31

This needs to be done:

- [x] support es6 classes
- [x] ~~support createClass()~~
- [x] support stateless components()
- [x] support all types from flow
  - [x] primitive types
  - [x] class types
  - [x] object signature
  - [x] literal types
  - [x] function signature
  - [x] callable-object/function-object signature
  - [x] maybe types
  - [x] union
  - [x] intersection
- [x] support inline type definitions
- [x] support declared type definitions
- [x] extract description from type
- [x] ~~handle imported type definitions (not really sure if we can do anything about that)~~
- [x] add tests
- [x] update ast-types to support Identifiers to TypeAlias
- [x] make code clean :)
- [x] adapt Readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,9 @@
 
 It uses [recast][] and [babylon][] to parse the source into an AST and provides methods to process this AST to extract the desired information. The output / return value is a JSON blob / JavaScript object.
 
-It provides a default implementation for React components defined via 
-`React.createClass`, [ES2015 class definitions][classes] or functions 
-(stateless components). These component definitions must follow certain 
+It provides a default implementation for React components defined via
+`React.createClass`, [ES2015 class definitions][classes] or functions
+(stateless components). These component definitions must follow certain
 guidelines in order to be analyzable (see below for more info).
 
 ## Install
@@ -41,11 +41,11 @@ Extract meta information from React components.
 If a directory is passed, it is recursively traversed.
 ```
 
-By default, `react-docgen` will look for the exported component created through 
-`React.createClass`, a class definition or a function (stateless component) in 
-each file. You can change that behavior with the `--resolver` option, which 
-either expects the name of a built-in resolver or a path to JavaScript module 
-exporting a resolver function. Have a look below for [more information about 
+By default, `react-docgen` will look for the exported component created through
+`React.createClass`, a class definition or a function (stateless component) in
+each file. You can change that behavior with the `--resolver` option, which
+either expects the name of a built-in resolver or a path to JavaScript module
+exporting a resolver function. Have a look below for [more information about
 resolvers](#resolver).
 
 Have a look at `example/` for an example of how to use the result to generate a
@@ -111,7 +111,9 @@ For example, while the `propTypesHandler` expects the prop types definition to b
 - `propTypes` must be an object literal or resolve to an object literal in the same file.
 - The `return` statement in `getDefaultProps` must contain an object literal.
 
-## Example
+## PropTypes
+
+### Example
 
 For the following component
 
@@ -156,7 +158,7 @@ module.exports = Component;
 
 we are getting this output:
 
-```
+```json
 {
   "props": {
     "foo": {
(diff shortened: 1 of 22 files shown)


fkling:
Nice, thank you! TBH, I had something different in mind with flow type support. I thought we could run flow directly and query it to get type information (though I don't know how much information we can get).

However, if we can do it this way and get equally good results, that's great!

danez:
I tried flow itself before, but the ast it delivers is nearly the same as the one from babylon. The only current drawback with babylon is that imported/required or types which are defines as libs for flow, cannot be resolved. But I also don't know how one would do that with flow. That might be a future addition if someone finds a way to do that.

I'm nearly done with all the codechanges. I probably might refactor smaller pieces, but I think there wont be large changes anymore.

I will now start updating the readme and add a section about flow, and what are the differences to propTypes. As flow offers more granular types, the results (except primitive types) are a little bit different.

fkling:
> I tried flow itself before, but the ast it delivers is nearly the same as the one from babylon.

I was thinking that flow can return us the type information already parsed, but since I haven't worked with it yet, I might be wrong.

Either way, really happy about this :) When you are done, I'd appreciate it if you could squash everything into a single commit.

danez:
Yes squashing everything was the plan. :)

danez:
Do you think that the flowHandler should be part of the default handlers?

README.md
```diff
@@ -233,8 +235,158 @@ The structure of the JSON blob / JavaScript object is as follows:
...
+            "name": "any"
+          }
+        ],
+        "raw": "Array<any>"
+      },
+      "required": true
+    },
+    "func": {
+      "type": {
+        "name": "functionshape",
```
danez: Im not sure about the name here. Maybe someone has a better idea?

README.md
```diff
@@ -233,8 +235,158 @@ The structure of the JSON blob / JavaScript object is as follows:
...
+          "__return__": {
+            "name": "void"
+          }
+        }
+      },
+      "required": false
+    },
+    "obj": {
+      "type": {
+        "name": "objectshape",
```
danez: Same here with the name. For propTypes this was just "shape"

phpnode:
This is fantastic, very much looking forward to using it. Is there anything I can help with to get this landed sooner?

danez:
Ok I am done from my side, as far as I can see.

The only remaining question is if the flowTypeHandler should be part of the default handlers (currently it is)? This probably depends on how you want to release this feature: With this enabled by default I would think it is more a 3.0 version (as it could generate different output than before for people), whereas if it is not enabled by default, then it could be 2.7. But thats up to you @fkling .

fkling:
Mmh, not sure, what would you prefer? Would it make sense to have a `--flow` option on the CLI which will use flow handlers, instead of standard ones, if they exist.

danez:
First of all flow- and propType-handlers don't conflict with eachother. Having both active is not a problem. The only "problem" that could arise is if people use both propTypes and flow in their components. In that case one of the handlers (depending on the order) would overwrite the other one.

But even if they don't conflict, I like the idea of having a `--flow` option in the cli, which adds the flow handler to the defaultHandlers if set and disables the propType handlers.

For the api i could export flowTypeHandlers, so people can easily add them to their handlers like

```js
import { defaultHandlers, flowTypeHandlers, parse } from 'react-docgen';

parse(code, [...defaultHandlers, ...flowTypeHandlers])
```

If thats all okay, I add that?

fkling:
> The only "problem" that could arise is if people use both propTypes and flow in their components. 

Right. I think (hope?) that this is not very common. What's more interesting is how often people use flow type annotation *and* `propTypes` in the same code base (in different components)(which would be the case for Facebook). Those might want to have both handlers enabled.

However, I think it would fine to have this option first (no major update needed) and then see if others need a more flexible setup.

danez:
Okay I added the cli option.

fkling:
Actually, here might be a better idea:

Lets simply add the flow prop types to a new key, e.g. `flowProps`, and add the handlers to the defaults. This way there won't be any conflicts and both annotations can be used (for static and runtime checks).

How does that sound?

danez:
But then you end up having defaultValues under `props` and the types + description under `flowProps` which to me is a little bit counterintuitive.

I was thinking about something similar: Having the flow type under `flowType` inside the props object, next to `type`. The only question then is what to do about the `description`, maybe `flowDescription` although that sounds confusing to me.

fkling:
> Having the flow type under `flowType` inside the props object, next to `type`. 

Seems fine too.

> The only question then is what to do about the `description`, maybe `flowDescription` although that sounds confusing to me.

Do you think people would add a description to both, the propTypes and the flow types in the same component? I think it's fine to assume that only one of them will be annotated. We could make it so that either the flow type description handler or the propType description handler doesn't set a description if there already is one set.

danez:
That seems a reasonable approach. The code could also warn if there are multiple descriptions found.

danez:
Ok i changed the flowType from `type` to `flowType`.

And both docgen handlers respect each other and don't update the description if already set, but trigger a warning.

I removed the cli option again and added flowTypeHandler and flowTypeDocHandler to defaultHandlers.

fkling:
Two comments inside (somehow I'm always adding the comments so that GH doesn't show them inline, but whatever).

If that's clarified I think we are good to go! Really awesome work, I'm sure lots of people will be happy about this, I certainly am :) Thank you!

danez:
You were right about the type<->flowType thing in the readme, I did not regenerate it with the latest changes. Done now.

danez:
And yes you are right, the required field is also set by the flowTypeHandler. although this is less critical than the type itself i would say. 

In case someone really uses both propTypes and flow, the type can be completely different, but the required field should be the same, And I think it would count as a bug in the components if you specify a prop as required with propTypes but not with flow.
Do you think it is worth adding a check if the other handler already added the required field?

fkling:
> Do you think it is worth adding a check if the other handler already added the required field?

I think its fine for now. If we encounter any problems, we can always update later. Everything looks really great! I'm only a little concerned regarding the warnings, since there is no way to disable them. This might be surprising if react-docgen is used as a module instead of the CLI.

OTOH, I think it would be valuable if there was some information about unresolved types. But I also don't want to block this on implementing a proper error handling feature.

Should we just omit the warnings for now?

danez:
Thats true. There is also a note in the readme about this unresolvable types.

danez:
Okay I removed them now.

fkling:
Thank you! I'm going to merge this now, and release a new version later today.

danez:
@fkling If there are bugs coming in, feel free to ping me.

benjamn:
Looking forward to that ast-types PR :)

Issue #49, opened 2016-01-11
Tests fail with recast >= 0.10.41


danez:
With recast >= 0.10.41 two tests are failing:

```
 FAIL  src/handlers/__tests__/componentDocblockHandler-test.js (11.578s)
● componentDocblockHandler › ClassDeclaration › decorators › it uses the component docblock if present
  - Expected: 'Decorator description' toBe: 'Component description'
        at Spec.eval (src/handlers/__tests__/componentDocblockHandler-test.js:124:43)
● componentDocblockHandler › ES6 named exports › Named class declaration export › decorators › it uses the component docblock if present
  - Expected: 'Decorator description' toBe: 'Component description'
```

I opened already an issue with recast benjamn/recast#245, but no response yet. Maybe you guys know why it changed.

fkling:
There have been some changes around how decorators are printed. I guess it may also have impacted comment attachment. Whether that's a bug or intended I don't know.

fkling:
Looks like a bug to me: The `comments` property contains the decorator comment, the `leadingComments` property contains the component comment: http://astexplorer.net/#/lBRfRefMun .

Could probably still fix this from my side.

PR #50, opened 2016-01-12
Do not compile tests and mocks to dist

danez:
To run the tests it is not needed to transpile them beforehand, and for the npm package
the mocks and tests are ignored anyway.
So I think it is safe to ignore them in the babel transformation.

The tests probably fail due to #49 .


fkling:
I fixed the test if you want to rebase!

danez:
Ok I rebased

PR #52, opened 2016-01-24
Parse description in propTypeHandler

vjeux:
This is needed for https://github.com/facebook/react-native/pull/5520

I found the line to add in `propDocBlockHandler.js`


vjeux:
Actually, this pull request is not needed. I just had to add `docgen.handlers.propDocBlockHandler` to the handler list :)

Issue #30, opened 2015-10-14
include displayName in docs


iamdustan:
Is there a reason this wasn’t already included? Having a displayName with each description would be beneficial for tools that are using docgen without having to re-read the js file to detect the display name again.

This wasn’t that big of a deal with `React.createClass` or `class extends React.Component` since those are pretty limited call sites, but the stateless function components can get a bit gnarly in trying to infer their display names. The detection for these are already pretty solid in the #28 branch.

fkling:
You mean why it isn't listed in the README? Probably just an oversight from my side.

iamdustan:
wait. does the displayName already appear in there?

iamdustan:
I don’t think it does... this is the result of console.logging

```
docgen.parse(fs.readFileSync('./my-component.js'))
```

```
{ description: '@private',
  props:
   { to: { type: [Object], required: true, description: '' },
     name: { type: [Object], required: true, description: '' } } }
```

fkling:
We do have https://github.com/reactjs/react-docgen/blob/master/src/handlers/displayNameHandler.js though.

fkling:
I can't repro this btw. The `displayName` shows up in the result. What does your `./my-component.js` look like?

iamdustan:
Hmm..

so I have stateless components not getting a `displayName` as well as

```
import React, {PropTypes, Component} from 'react';
class Tabs extends Component {
  ....
}

export default Tabs;
```

fkling:
Oh, the `displayName` handler only handles explicit `displayName` properties. You want it to also return the inferred name from the class or function name, right?

iamdustan:
:+1:

...does babel still follow that approach for displayName handling?

mik01aj:
@fkling, yes! (I know the question wasn't to me, but I'd love to have it this way :))

fkling:
I don't know, will have to look into this. Seems reasonable to me though. PRs are also welcone ;)

iamdustan:
I vote @mik01aj for those one :)

AlanFoster:
Babel will generate the display name for you when it is not supplied explicitly. 

:+1: to implementing a similar inferrance rule to react-docgen

AlanFoster:
@iamdustan Would this be something that is feasible to implement?

mik01aj:
Maybe you could just use this plugin: https://github.com/gajus/babel-plugin-react-display-name

NogsMPLS:
[...]

PR #59, opened 2016-03-01
infer displayName when not defined

NogsMPLS:
For helping out with #30. 

Added tests and passes tests on my end for es6 class with no static displayName attached and also stateless component.

Would like a second pair of eyes in `src/handlers/displayNameHandler.js` on `line 34`. I'm not sure if I like that solution using `[0]`, but not sure how to test around it.


src/handlers/displayNameHandler.js
```diff
@@ -23,12 +23,16 @@ export default function displayNameHandler(
...
-    return;
+
+  if (displayNamePath) {
+    displayNamePath = resolveToValue(displayNamePath);
+    if (!displayNamePath || !types.Literal.check(displayNamePath.node)) {
+      return;
+    }
+    displayNamePath = displayNamePath.node.value
+  } else if (!displayNamePath) {
+    displayNamePath = path.node.id ? path.node.id.name : path.node.declarations[0].id.name;
```
NogsMPLS: Would like a second pair of eyes on this, specifically around ```path.node.declarations[0].id.name```. I'm not sure if I like it.

fkling:
Thanks for working on this! I will add some comments inline.

src/handlers/__tests__/displayNameHandler-test.js
```diff
@@ -70,17 +70,30 @@ describe('defaultPropsHandler', () => {
...
+    expect(documentation.displayName).toBe('Foo');
+  });
+
+  it('infers the displayName with stateless functional component', () => {
+    var definition = expression('({displayName: "FooBar"})');
+    displayNameHandler(documentation, definition);
+    expect(documentation.displayName).toBe('FooBar');
+
+    definition = statement(`
+      var Foo = () => {
```
fkling: Unfortunately this is not useful test, because the `displayNameHandler` function will never be passed a *VariableDeclaration*.

If you look at `getMemberValuePath`, you can see the node types which represent component definitions:

```
var LOOKUP_METHOD = {
  [types.ArrowFunctionExpression.name]: getMemberExpressionValuePath,
  [types.FunctionExpression.name]: getMemberExpressionValuePath,
  [types.FunctionDeclaration.name]: getMemberExpressionValuePath,
  [types.VariableDeclaration.name]: getMemberExpressionValuePath,
  [types.ObjectExpression.name]: getPropertyValuePath,
  [types.ClassDeclaration.name]: getClassMemberValuePath,
  [types.ClassExpression.name]: getClassMemberValuePath,
};
```

For some of them, inferring the name is easy: You can simple take the `id` of named classes and functions.

For all others, you need to traverse up the tree to find and assignment or export and inspect that one to get the name.

Here is how `getMemberExpressionValuePath` does it: https://github.com/reactjs/react-docgen/blob/9042c91d6b9e787f4fa49a96b7be0babb256fd3c/src/utils/getMemberExpressionValuePath.js#L36-L45

src/handlers/displayNameHandler.js
fkling: See my comment above :) (which should make this a non-issue)

src/handlers/__tests__/displayNameHandler-test.js
NogsMPLS: Okay so, I still want to be able to infer the name and assign the as the displayName of stateless functional component.

What would you think about exporting ```resolveName()``` from ```/getMemberExpressionValuePath.js```, importing it into ```displayNameHandler``` and using that to grab the variable definition display names?

NogsMPLS:
blah, fail test. ill check

fkling:
> Would breaking resolveName into separate file be cool?

Sure! Can you also update the tests to verify whether these cases are covered?

- Function expression: `(function Component() {})`
- Named exports:
  - `export var Foo = function() {}`
  - `module.exports = function() {}`

It was also be great to get some test coverage for `React.class(...)` component definitions. This might require some changes to the way the tests are set up. Have a look at https://github.com/reactjs/react-docgen/blob/master/src/handlers/__tests__/componentDocblockHandler-test.js for how different component definitions could be tested easily.

NogsMPLS:
You're talking about mainly updating the ```displayNameHandler-test.js``` to handle those use cases right?

fkling:
Yes.

NogsMPLS:
Almost got this, named exports giving me a little trouble, but I think I should have something tomorrow.

NogsMPLS:
The biggest issue I'm running into with Named Exports has to do with `Node` vs. `NodePath`.

Basically, inside `getMemberValuePath.js` I can get the declaration of the export (i.e. `VariableDeclartion`/`ClassDeclaration`/`FunctionDeclaration`) by doing something like:

```javascript
//checks if named export, then redefines the componentDefinition to be the export's declaration
  if (componentDefinition.node.type === types.ExportNamedDeclaration.name) {
    componentDefinition = componentDefinition.get('declaration').value;
}
```

However, once I grab it and redefine `componentDefinition` to the new declaration, `componentDefinition` stops being a `NodePath` and starts being a `Node`, which screws with about 4 other files that end up using `path.node`, which, as far as I can tell, only exists for `NodePath` type.

I think I can make it work all around, but it's gonna be a bigger refactor than just 1 file.

fkling:
After [looking at the AST](http://astexplorer.net/#/oraLq8MoAE), it seems named exports should actually be covered by variable declarations, no?

Regarding `exports.name = ...` it seems you would need to look whether the parent is an assignment expression to `exports.X` and grab the property name.

`getMemberValuePath` is for getting keys from the component definition (i.e. `propTypes`, `defaultProps`, etc), it wouldn't help you finding information "outside" the component definition.

If you have something new to share, I'm happy to take a look.

NogsMPLS:
Oh neat, didn't know that ast tool existed. (I'm still pretty new to parsing the AST)

well, so, first thing I did was just add a test to `displayNamedHandler-test.js`, that was simply:

```javascript
it('infers the displayName with named export', () => {
    var definition = statement(`
      export var Foo = function() {}
    `);
    displayNameHandler(documentation, definition);
    expect(documentation.displayName).toBe('Foo');
  });
```
But of course, that throws an error because the definition type is `ExportNamedDeclaration`, which `getmemberValuePath.js` doesn't like.

So my line of reasoning was to go one level deeper into the `ExportNamedDeclaration` AST object which things like `VariableDeclaration` is defined and use that, but I'm ran into `path.node` not existing issues.

NogsMPLS:
i expect this latest to trip travis-ci. just pushing so i have it up there for now.

NogsMPLS:
Was able to get infer name on es6 exports to work on classes, function expressions and variable declarations. (though I had to do some stuff that feels very clunky for variable declaration)

now to get ```exports.name``` working and we should be there

NogsMPLS:
Got a version of infering displayName on commonJS exports working.

Not sure if I like the solution right now.

Need to add more tests around exports and attaching displayName vs. inferring it and make sure the more explicit attaching of displayName takes precedent.

src/handlers/displayNameHandler.js
```diff
@@ -13,22 +13,65 @@
...
+    if (
+        declaration.type === types.ClassDeclaration.name ||
+        declaration.type === types.FunctionDeclaration.name ||
+        declaration.type === types.FunctionExpression.name
+      ) {
+      declarationPath = resolveExportDeclaration(path)[0];
+
+    } else if (declaration.type === types.VariableDeclaration.name) {
+
+      declarationPath = resolveExportDeclaration(path)[0].parentPath.parentPath.parentPath;
```
NogsMPLS: I don't really like this solution, but I'm not sure of how else to get this value while keeping 'NodePath' type intact.

NogsMPLS:
Looking at stuff over against today, I decided it's probably best to keep current behavior in regards to an example like:
```javascript
export class Foo {
   static displayName = foo.bar;
}
```
Initially, I was going to try and have it infer the class name as the displayName in that case, but I think the better behavior might be to keep it empty returning as it currently is.

All tests are passing, would like a review, especially on how I'm handling immediately exported variable declarations.

Other than that though, I think this satisfies #30 as far as I can tell.

src/handlers/__tests__/displayNameHandler-test.js
```diff
@@ -70,17 +70,161 @@ describe('defaultPropsHandler', () => {
...
+    var definition = statement(`
+      (function FooBar() {})
+    `);
+    displayNameHandler(documentation, definition);
+    expect(documentation.displayName).toBe('FooBar');
+  });
+
+});
+
+describe('defaultPropsHandler with ES6 Exports', () => {
```
fkling: I guess this should be "dispayNameHandler".

fkling:
I'll take a closer look again today, thanks!

NogsMPLS:
hmmm....tests passed on stable and node 5 but failed on node 4

Looks like it's failing on a file this pull request doesn't touch. Seems like it's maybe a fluke?

NogsMPLS:
looks like this is what is keeping #97 from passing as well.

I did a bit of digging and found a jest issue mentioning this:

https://github.com/facebook/jest/issues/595

This seems to be an intermittent fail from what I can tell.

perhaps updating jest from `^12.1.1` to `^14.0.1` might fix, but going up 2 majors should probably be its own pull request.

If you think getting jest updated is important, I can work and get a PR in for that.

fkling:
We can at least give it a try yes!

NogsMPLS:
Oh also, I 100% completely understand if this branch/feature isn't something you'd want merged into the main trunk. It solves a problem for my purposes though, so I've been linking to a fork of this for my projects.

Along those lines, I'd like to ask your permission for me to create a npm package with this feature included in it so I can stop linking directly to github in a few of my projects.

ZauberNerd:
@NogsMPLS I'm also very interested in using this. If it is not going into upstream, could you create a separate npm package for this handler?
I documented a little bit how I created a custom handler in this issue: https://github.com/reactjs/react-docgen/issues/115 
As an example you could take a look at these two packages: [nerdlabs/react-docgen-imports-handler](https://github.com/nerdlabs/react-docgen-imports-handler) and [nerdlabs/react-docgen-exports-handler](https://github.com/nerdlabs/react-docgen-exports-handler)
As Felix seems to be very busy at the moment (see linked issue above) I'd probably just go ahead and create an npm package for this. And if this should be in upstream at a later stage you could still open a new PR and deprecate your package?

NogsMPLS:
Ah, interesting, I didn't see that post about future of react-docgen.

yeah I can look into making this into a custom handler.

fkling:
>  I'd probably just go ahead and create an npm package for this.

yeah, that would be the best solution for now. I'm happy to merge any fixes that would need to be done to core utilities though.

ZauberNerd:
@NogsMPLS So, I've had some spare time and went ahead and created a displayName handler. Would that be covering all of your use cases too? https://github.com/nerdlabs/react-docgen-displayname-handler

NogsMPLS:
@ZauberNerd nice! Ill check it out sometime this week and let you know

danny-andrews:
Is this PR died? It would be really nice to have this feature!

NogsMPLS:
@danny-andrews I believe the display-name-handler that @ZauberNerd created does a decent job of covering what this PR was meant to do.

I know of at least 1 pretty popular repo that is using that display-name-handler plugin: https://github.com/styleguidist/react-styleguidist

I am going to go ahead and close this PR with the advice that people look into using that plugin instead of anything in this PR specifically.

sapegin:
Yeah, we’re using display-name-handler in React Styleguidist and it works perfectly!

danny-andrews:
Thanks!

PR #60, opened 2016-03-02
ignore Webstorm .idea directory

NogsMPLS:
just a simple git ignore to clear up git status message for people using Webstorm or other IntelliJ IDE's


fkling:
In general I hope people add their environment specific configuration to their global or user `.gitignore` file, but why not :)

PR #77, opened 2016-05-18
Fix findExportedComponentDefinition unit tests

rtsao:
This suite of tests was *always* being *erroneously* marked as passing due to the `describe.only` on line 486. Even adding `expect(false).toBeTruthy();` to any test would result in this suite still passing.

This PR also fixes the broken tests, including a syntax error in one of the source fixtures.

Note: tests on `master` are currently failing on Node 6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,4 +2,4 @@ sudo: false
 language: node_js
 node_js:
   - 4
-  - stable
+  - 5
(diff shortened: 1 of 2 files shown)


facebook-github-bot-1:
Thank you for your pull request and welcome to our community. We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file. In order for us to review and merge your code, please sign up at https://code.facebook.com/cla - and if you have received this in error or have any questions, please drop us a line at cla@fb.com. Thanks!

fkling:
Thanks for catching that! I will look into the Node6 issues...

PR #81, opened 2016-05-31
Fix tests on node 6

danez:
The tests were failing on node v6 because jest 0.8 and node 6 were not compatible. What I did is updating jest to 12 and then the journey began as jest 12 only supports babel 6.

In total this was done:

* Update jest to latest version
* Update babel-jest to latest version
* Upgrading babel to latest version (6.9) for jest
* Rewrite the custom matcher to conform to jasmine2 specs
* Fix deprecated jest apis (autoMockOff() => disableAutomock(), dontMock() => unmock())
* Babel 6 is not exporting the default export as module.export, so `.default` had to be added in the tests and the binary in order for the require() to work
* As I was already updating stuff I also switched from `cross-spawn-async` to `cross-spawn`, as the first one is deprecated.
* changed from `rm -rf` to `rimraf` in `npm build` for windows support.
* Update recast to 0.11.x as the only change from 0.10 to 0.11 was the default parser switched from esprima-fb to esprima, but react-docgen is using babylon anyway.
* Some tests for stateless components in `findExportedComponentDefinition` were broken and not run as there was a `desribe.only` which made the old jest silently ignoring these tests.

(Babylon is still at version 5.8, if there is ever the need to update babylon then this is complete separate story)

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,11 @@
 {
-  "optional": ["runtime"]
+  "presets": [
+    "es2015",
+    "stage-1"
+  ],
+  "plugins": [
+    "syntax-flow",
+    "transform-flow-strip-types",
+    "transform-runtime"
+  ]
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,3 +3,4 @@ language: node_js
 node_js:
   - 4
   - 5
+  - stable
--- a/bin/__tests__/react-docgen-test.js
+++ b/bin/__tests__/react-docgen-test.js
@@ -8,18 +8,19 @@
  *
  */
 
-/*global jasmine, jest, describe, pit, expect, afterEach*/
+/*global jasmine, jest, describe, it, expect, afterEach*/
 
-// Increase default timeout (5000ms) for Travis
-jasmine.getEnv().defaultTimeoutInterval = 10000;
+const TEST_TIMEOUT = 120000;
 
-jest.autoMockOff();
+jasmine.DEFAULT_TIMEOUT_INTERVAL = TEST_TIMEOUT;
+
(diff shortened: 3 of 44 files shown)


danez:
Travis is too slow to run the tests for the cli and jest does not forward the test timeout currently. Waiting for an answer here facebook/jest#1090

fkling:
Thanks for all your help and sorry for neglecting rect-docgen recently. I'm really swamped with other stuff... lets see how quickly there will be a new jest version.

fkling:
The tests still fail for Node v4 in Travis CI. Seems to be only a timeout error though. It seems to pass fine locally. Could you try updating the timeout to 30 sec?

danez:
There is still no new version of jest, and I couldn't figure out how to change the timeout without the fix.

fkling:
Oh, I thought they would have released one by now :(

I just realized that I already updated jscodeshift to work with jasmine2. There I am using

```
jasmine.DEFAULT_TIMEOUT_INTERVAL = 30000;
```

That should work.

danez:
I increased the timeout now to 120s as everything below was not working. The tests with node4 and npm2 are terrible slow on travis.

fkling:
oh my :-/ Thank you!

fkling:
Merged with 6a655bd613109c622900c8c7af81397db2347dcf .

fkling:
Tests still fail for Node v4 from time to time, but they work fine locally so ¯\_(ツ)_/¯.

danez:
Yeah I really can't figure out why the tests should by 100 times slower. If i use npm 2 locally it is a little bit slower, but not thaaat slow.

PR #83, opened 2016-06-14
Change typealias name add param optional

caabernathy:
To support some proposed React Native doc updates, here are two changes:

1. Expose the type name, ex: the code below would like to see type.name set to StatusBarStyle instead of $Enum:

        export type StatusBarStyle = $Enum<{
             ...
        }

2. Expose whether a method's parameter is optional or not. Currently this isn't brought out.

### DIFF
--- a/src/utils/getMethodDocumentation.js
+++ b/src/utils/getMethodDocumentation.js
@@ -14,10 +14,14 @@ import getFlowType from './getFlowType';
 import getParameterName from './getParameterName';
 import getPropertyName from './getPropertyName';
 import getTypeAnnotation from './getTypeAnnotation';
+import recast from 'recast';
+
+const { types: { namedTypes: types } } = recast;
 
 type MethodParameter = {
   name: string;
   type?: ?FlowTypeDescriptor;
+  optional?: boolean;
 };
 
 type MethodReturn = {
@@ -42,10 +46,14 @@ function getMethodParamsDoc(methodPath, jsDoc) {
     const typePath = getTypeAnnotation(paramPath);
     if (typePath) {
       type = getFlowType(typePath);
+      if (types.GenericTypeAnnotation.check(typePath.node)) {
+        type.alias = typePath.node.id.name;
+      }
     }
 
     const param = {
       name: getParameterName(paramPath),
+      optional: paramPath.node.optional,
       type,
     };
 


caabernathy:
FYI, I see the same Travis CI failures when I run tests with node v6.2.1 and un-changed code from master. So I don't believe my changes introduced any new errors.

danez:
Yes the tests are currently failing on node 6. #81 will fix that

src/utils/getMethodDocumentation.js
```diff
@@ -42,10 +46,14 @@ function getMethodParamsDoc(methodPath, jsDoc) {
     const typePath = getTypeAnnotation(paramPath);
     if (typePath) {
       type = getFlowType(typePath);
+      if (types.GenericTypeAnnotation.check(typePath.node)) {
+        type.name = typePath.node.id.name;
+      }
```
danez: This check might be better in getFlowType. I can check, I couldn't remember why it is currently not working

danez: I'm not sure if that is always a wanted behavior. If you have code that looks like this:
```
class Foo {
          foo(bar: StatusBarStyle) {}
        }
        
        type StatusBarStyle = Array<{blub:string}>
```

You might want to know that the type is Array and not StatusBarStyle. Even with $Enum I might tend to say that $Enum is the type-name and not the title of the alias.

Maybe it would be good in the case of Aliased types to always have both, the alias and the resulting type in the doc. Because both might be super useful to display.

caabernathy:
@janicduplessis or @fkling thoughts on this PR? I was looking for guidance in light of a React Native doc lockdown scheduled for next week.

janicduplessis:
I think we should keep name as it is and add this as another property like `alias` or whatever name makes more sense :)

caabernathy:
Great idea, I was thinking the same but wasn't sure. I'll make that change.

caabernathy:
I'd appreciate another look @janicduplessis and @fkling.

janicduplessis:
Looks good to me, there is an issue with the PR though, it includes a bunch of unrelated commits.

fkling:
Yeah, you need to rebase your changes on top of current master. But I can probably do this too if you don't get the chance.

caabernathy:
Ok, rebased.

Issue #82, opened 2016-06-07
Dependency detection


khankuan:
Hi, wondered if it is possible to parse the jsx and obtain all Tag names (maybe exclude native tags) to form the dependency list for the component.

Cheers :)

fkling:
This should be possible. You can define your own handler which traverse the AST of the `render` (and any other method you are interested in) and look for JSX elements.

Do you have a more specific about that?

khankuan:
Ic. I saw that there is the `composes` key in the output. Is that in used at the moment?

fkling:
Yes. This is populated with the component names when you do

```
propTypes: {
   ...SomeOtherComponent.propTypes,
}
```

But if you want to add to that from your own handler, that's fine I guess. It's your code after all :)

PR #93, opened 2016-06-27
Feature/dependencies

khankuan:
Initial implementation for #82. Open to any feedbacks to get the feature available as soon as possible :)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -4,10 +4,13 @@
     "comma-dangle": [1, "always-multiline"],
     "no-underscore-dangle": 0,
     "quotes": [2, "single", "avoid-escape"],
-    "strict": 0
+    "strict": 0,
+    "no-unused-vars": 2,
+    "no-undef": 2
   },
   "env": {
-    "node": true
+    "node": true,
+    "es6": true
   },
   "globals": {
     "ASTNode": true,
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,4 +1,0 @@
-{
-  "node": true,
-  "esnext": true
-}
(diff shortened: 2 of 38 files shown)


fkling:
So, what would be the difference between `composes` and `dependencies` now?

Do you think this handler needs to be part of react-docgen core? I'm hesitant to include any more handlers. I think this could every well be a plugin that people can use if they want to. I think we'd only need to make it easier to include handlers via the command line. What do you think?

khankuan:
I feel that composes feels like it is targeted for the case:
```
<div>
  <A { ...propsA } />
  <B { ...propsB } />
</div>
```
Whereas dependencies does not really imply any propTypes and focus on the components that are referenced in the render method. I felt that it is unnecessary to enforce that composition is done via propTypes with the spread operator. In some cases, there is no need to declare the propTypes as well such as:
```
<div>
  My Content
  <Divider />
  My Second Content
</div>
```
Also, the proposed method also includes native tags that can be useful as well.

I thought that the composition is one of the fundamental strengths of React and documenting component dependencies is an obvious benefit to be included by default. I would think that dependencies might be more useful than composes for beginners. Is there a list of plugins that are build on top of react-docgen? Personally, I would like having the handler here with all the build and test processes already well crafted and we have more radar on future regressions.

Another consideration is to combine composes and dependencies but it might dilute the meaning of composes for devs that are already using it.

Cheers :)

khankuan:
Hey, wondered if the PR could be reconsidered. Otherwise, would it be possible to cherry pick just the changes on this file: https://github.com/khankuan/react-docgen/blob/4dfb0fbc5d9ff43df9fdc5f1dfd9f6797f129728/src/utils/resolveToModule.js

fkling:
You probably don't need this anymore but I update `resolveToModule` in 2c6e55aa838c1cb9a6755711a17ca4f108cbdf13 😉

PR #95, opened 2016-07-05
fix jsdoc @param union type support & @returns * all type support

dickeylth:


dickeylth:
Two weeks gone, is there anyone maintaining this package's upgrade? Need this pr to be merged badly!

fkling:
Sorry, I've been on vacation for exactly two weeks :)

src/utils/__tests__/parseJsDoc-test.js
```diff
@@ -64,6 +64,21 @@ describe('parseJsDoc', () => {
...
+    it('extracts jsdoc union type param', () => {
+      const docblock = `
+        @param {string|Object} bar
+      `;
+      expect(parseJsDoc(docblock)).toEqual({
+        description: null,
+        returns: null,
+        params: [{
+          name: 'bar',
+          type: {name: 'string|Object'},
```
fkling: I'm wondering if this shouldn't rather be something like `{name: 'union', value: ['string', 'Object']}`. What do you think @janicduplessis?

janicduplessis: I agree we should keep this as an object since it's easier to use after. We should try to have a similar output object as the flow type definitions.
```
/**
 * @param test Test
 */
function (test: string | Object) {}
```
and
```
/**
 * @param {string | Object} test Test
 */
function (test) {}
```
should result in the same thing.

You could take a look at what the object we return for flow types look like and return something similar.

dickeylth: OK, I'll update the commit soon.

dickeylth:
Another week gone, so when will this to be merged and when will next version to be published?

fkling:
Sorry for the slow progress here, the last couple of weeks just had been really busy (I was traveling most of the time).

If we want to stay compatible with flow types, I think `*` should be mapped to `mixed` or `any`.

dickeylth:
OK, thank you for your suggestion, I've updated the commit~

danez:
👍

PR #97, opened 2016-07-08
Refactor for flow intersection and union Props

palange:
This follows up to the flow support added by @danez

Refactor to resolve, when possible, a list of props
from a flow intersection type and to throw an error
when encountering a Props of union type.

This change addresses a bug found in the flowTypeDocBlockHandler
while a props type that is an IntersectionTypeAnnotation or a
UnionTypeAnnotation.

Union types throw an error, rather than generating misleading documentation.
When given a union type, flow will enforce that props conform to only one of
the union-ed types. The documentation format, which lists a single set of props
can't express this.

### DIFF
--- a/src/handlers/flowTypeHandler.js
+++ b/src/handlers/flowTypeHandler.js
@@ -14,27 +14,23 @@ import type Documentation from '../Documentation';
 
 import getFlowType from '../utils/getFlowType';
 import getPropertyName from '../utils/getPropertyName';
-import getFlowTypeFromReactComponent from '../utils/getFlowTypeFromReactComponent';
+import getFlowTypeFromReactComponent, {
+  applyToFlowTypeProperties,
+}  from '../utils/getFlowTypeFromReactComponent';
+import isUnreachableFlowType from '../utils/isUnreachableFlowType';
+import recast from 'recast';
+import resolveToValue from '../utils/resolveToValue';
+import { getDocblock } from '../utils/docblock';
+
+
+var {types: {namedTypes: types}} = recast;
 
 function setPropDescriptor(documentation: Documentation, path: NodePath): void {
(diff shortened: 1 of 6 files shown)


palange:
Updated commit to fix lint errors

src/utils/getFlowTypeFromReactComponent.js
```diff
@@ -51,7 +51,50 @@ export default (path: NodePath): ?NodePath  => {
...
+      propertyPath => callback(propertyPath)
+    );
+  } else if (path.node.type == 'IntersectionTypeAnnotation') {
+    path.get('types').each(
+      typesPath => applyToFlowTypeProperties(typesPath, callback)
+    );
+  } else if (path.node.type == 'UnionTypeAnnotation') {
+    // The react-docgen output format does not currently allow
+    // for the expression of union types
+    throw new TypeError("react-docgen doesn't support Props of union types");
```
danez: I agree now, that generating props from union types is more tricky and a bad idea like it was before, but imho it would be better to just ignore the type then instead of throwing an exception. People should not be forced to restructure their code just to get the docs working.

src/handlers/flowTypeHandler.js
```diff
@@ -14,30 +14,26 @@ import type Documentation from '../Documentation';
...
+import getFlowTypeFromReactComponent, {
+  applyToFlowTypeProperties,
+}  from '../utils/getFlowTypeFromReactComponent';
+import isUnreachableFlowType from '../utils/isUnreachableFlowType';
+import recast from 'recast';
+import resolveToValue from '../utils/resolveToValue';
+import { getDocblock } from '../utils/docblock';
+
+
+var {types: {namedTypes: types}} = recast;
```
danez: I think most of this imports above are unused.

danez:
This has been merged with #118.
Thanks a lot.

PR #99, opened 2016-07-22
Check for import declaration type when getting default values, resolves #63

### LINKED ISSUE #63: Imported Default Prop Types
I seem to be having an issue with the following code -

```js
import React from 'react';
import DefaultChildComponent from './default-child-component.jsx';

const ParentComponent = React.createClass({
  getDefaultProps() {
    return {
      ChildComponent: DefaultChildComponent
    }
  },

  render() {
    return (
      <div>
        <this.props.ChildComponent foo={true} />
      </div>
    )
  }
});

export default ParentComponent;
```

This triggers the following error -

```shell
    throw (0, _pluginError2['default'])({
    ^
Error: [{type: ImportDefaultSpecifier, start: 34, end: 55, loc: [object Object], local: [object Object], id: null, name: null}] does not match type Printable
    at Type.Tp.assert (/Users/alanfoster/Documents/Shopkeep/backoffice/node_modules/ast-types/lib/types.js:56:15)
    at genericPrintNoParens (/Users/alanfoster/Documents/Shopkeep/backoffice/node_modules/recast/lib/printer.js:162:26)
    at genericPrint (/Users/alanfoster/Documents/Shopkeep/backoffice/node_modules/recast/lib/printer.js:148:33)
    at printRootGenerically (/Users/alanfoster/Documents/Shopkeep/backoffice/node_modules/recast/lib/printer.js:94:16)
    at maybeReprint (/Users/alanfoster/Documents/Shopkeep/backoffice/node_modules/recast/lib/printer.js:88:16)
    at print (/Users/alanfoster/Documents/Shopkeep/backoffice/node_modules/recast/lib/printer.js:75:29)
    at exports.printComments (/Users/alanfoster/Documents/Shopkeep/backoffice/node_modules/recast/lib/comments.js:324:22)
    at printWithComments (/Users/alanfoster/Documents/Shopkeep/backoffice/node_modules/recast/lib/printer.js:61:16)
    at print (/Users/alanfoster/Documents/Shopkeep/backoffice/node_modules/recast/lib/printer.js:66:20)
    at Printer.print (/Users/alanfoster/Documents/Shopkeep/backoffice/node_modules/recast/lib/printer.js:107:21)
```

nathanmarks:
A common pattern in libraries is to set a default prop value as an imported component, for eg. a transition component for a dialog.

#63 has an example + a workaround (not the most elegant however, involves using a functional component to return the default component).

Right now, `react-docgen` throws an error on components implementing this pattern as `ImportDefaultDeclaration` is not a printable type.

This PR adds a conditional after the `resolveToValue(path)` call in `getDefaultValue(path)` to check if the node is an import declaration. If so, it returns the `.name` property of the original node (so the function/component name).

### DIFF
--- a/src/handlers/__tests__/defaultPropsHandler-test.js
+++ b/src/handlers/__tests__/defaultPropsHandler-test.js
@@ -86,6 +86,27 @@ describe('defaultPropsHandler', () => {
       `;
       test(parse(src).get('body', 0));
     });
+
+    it('should find prop default values that are imported variables', () => {
+      var src = `
+        import ImportedComponent from './ImportedComponent';
+
+        class Foo {
+          static defaultProps = {
+            foo: ImportedComponent,
+          };
+        }
+      `;
+      defaultPropsHandler(documentation, parse(src).get('body', 1));
+      expect(documentation.descriptors).toEqual({
+        foo: {
+          defaultValue: {
+            value: 'ImportedComponent',
+            computed: true,
+          },
+        },
+      });
+    });
   });
 
   describe('ClassExpression with static defaultProps', () => {
--- a/src/handlers/defaultPropsHandler.js
+++ b/src/handlers/defaultPropsHandler.js
@@ -27,8 +27,12 @@ function getDefaultValue(path) {
     defaultValue = node.raw;
   } else {
     path = resolveToValue(path);
-    node = path.node;
-    defaultValue = printValue(path);
+    if (types.ImportDeclaration.check(path.node)) {
+      defaultValue = node.name;
+    } else {
+      node = path.node;
+      defaultValue = printValue(path);
+    }
   }
   if (typeof defaultValue !== 'undefined') {
     return {


brentschroeter:
+1: Code looks good. I've been including a similar handler manually in my application and would love to see Nathan's fix made standard.

fkling:
yeah, this looks good to me. I was wondering if it makes sense to include some additional information about the origin of the value (e.g. that it is the default import of the `./ImportedComponent` module), but I'm not sure where / how.

nathanmarks:
@fkling We can definitely provide some more information.

We know whether it was a default or a named import, and we know the require path.

Are you looking to provide information from the imported module itself? Or are you looking for something more specific about the import statement, ie: `default export from module './ImportedComponent`, `export NamedExport from module './ImportedComponent'`?

fkling:
> Are you looking to provide information from the imported module itself? 

No. This has come up a few times and maybe we are able to resolve and inspect dependencies at some point, but I was only thinking about the import statements itself.

In particular, if someone generates the docs, I can imagine they might want to link the source of the module or some other part of the page that provides more detailed information about `ImportedComponent`, and for that they probably need the path at least.

nathanmarks:
We can provide some additional information in the returned object.

Let me know **exactly** what you'd like to see, and I'll fix it up.

levithomason:
@fkling any way we can get this merged for now?  possibly add more info about the module later?  It is currently breaking our doc build.

fkling:
@levithomason, sure! I will try to release a new version today or tomorrow.

levithomason:
Super fast, thanks much!

PR #101, opened 2016-07-25
Tolerate non-object flow types.

pvolok:
React-docgen fails if props flow type is not an object.


fkling:
Seems reasonable to me. @danez ?

danez:
I thought I fixed that in #78, at least intersection and union types, which is used here in the test. It might still fail if one does `type Props = string;`. 
Give me a minute.

danez:
I created a new PR that generalizes the fix done in #78 so it can also be applied to the `flowTypeDocBlockHandler`.

#78 was only fixing the `flowTypeHandler`, wither I forgot the other Handler or it was added afterwards.

PR #102, opened 2016-07-25
added exclude cli parameter to exclude files by pattern

wallaroo:
hi,
i added a cli parameter 'exclude' to confugure some filename patterns from parsing.
the default is set to ['**index.js**'] because is a common use to organize components in indexed namespaces
Es.
```
/package.json
/app.js
/components/
/components/namespace1/index.js
/components/namespace1/Component1.js
/components/namespace1/Component2.js

```
where index.js is:
```
export Component1 from './Component1'
export Component2 from './Component2'

```
and app.js imports these components with:
`import {Component1, Component2} from './components/namespace1'`

i know that someone use the pattern `/components/ComponentName/index.js` (i don't like this) so... maybe is better to set [] as default...

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Options:
    -o FILE, --out FILE   store extracted information in FILE
    --pretty              pretty print JSON
    -x, --extension       File extensions to consider. Repeat to define multiple extensions. Default:  [js,jsx]
-   -e, --exclude         Filename pattern to exclude. Default:  [index.js]
+   -e, --exclude         Filename pattern to exclude. Default:  []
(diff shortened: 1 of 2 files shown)


fkling:
That's great, thank you! Yeah, I think we should leave the default empty, otherwise this would be a breaking change.

dptoot:
Definitely looking forward to this being added.  Any chance this can get added into the next release?

fkling:
@dptoot , @wallaroo just needs to remove the default value and this should be good to go 😃

wallaroo:
sorry for the late React...ion  ;-)

PR #104, opened 2016-08-02
update jest to 14

NogsMPLS:
updated jest, ran `npm run test` and tests passed on my OSX machine, so from what I can tell no breaking changes for this repo.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
   "devDependencies": {
     "babel-cli": "^6.9.0",
     "babel-eslint": "^6.0.4",
-    "babel-jest": "^12.1.0",
+    "babel-jest": "^14.1.0",
     "babel-plugin-syntax-flow": "^6.8.0",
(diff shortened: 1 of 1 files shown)


NogsMPLS:
Hopefully this also fixes https://github.com/reactjs/react-docgen/pull/59#issuecomment-236194249

🙏

NogsMPLS:
😭 😭 😭 😭 

node 4 env failed, just like the other PRs

fkling:
😞

fkling:
I guess I'll have to investigate a bit 🔍  ...

danez:
Dang didn't see there is already a PR for this (#106). Seems the tests on npm@2 sometimes fail, because they take too long. I was wondering if jest is maybe compiling the node_modules, but they should be ignored by default. So its probably babel that is slow and makes the worker crash/timeout.

danez:
this was fixed with #106

PR #105, opened 2016-08-02
Fix #101 - Do not fail on invalid flowTypes

danez:
Extract algorithm for finding props in flowType to utils from
flowTypeHandler so that it can be also used in flowTypeDocBlockHandler.

The same error that is described in #101 was already fixed #78 but then not carried on to the DocBlock Handler as well.


danez:
damn, I should really look at other commits before creating a new one.
Actually I like #97 better. Gonna add some comments to it and close this one.

fkling:
Thanks @danez!

PR #111, opened 2016-08-09
Verify that the MemberExpression that is being searched for belongs to the correct object

ZauberNerd:
When searching for a MemberExpresion (i.e. "propTypes") that is defined
on an object (ArrowFunctionExpression, FunctionExpression,
FunctionDeclaration or VariableDeclaration), take the local Identifier
into account to find the member of the correct object.

Before:
When using multiple components per file, it would for every component
take the propTypes of the last component encountered.
```javascript
function Component1() { return (<i />); }
Component1.propTypes = { a: React.PropTypes.string };
function Component2() { return (<b />); }
Component2.propTypes = { b: React.PropTypes.number };
```

After:
Now it checks whether the property of the MemberExpression actually
belongs to the correct object.


fkling:
Can't believe how this wasn't noticed earlier. Thank you!!

PR #112, opened 2016-08-09
Find nested MemberExpressions / resolve commonJS properties

ZauberNerd:
When an object doesn't have a "local identifier", for example because
it is directly assigned to "module.exports", we still want to be able
to find the member that we are looking for (i.e. "propTypes").

Before:
When directly assigning a React Component to "(module.)exports", the
correct members (i.e. "propTypes") could not be found.

``` javascript
module.exports = function() { return (<i />); };
module.exports.propTypes = { a: PropTypes.string };
```

or

``` javascript
module.exports.Foo = function() { return (<i />); };
module.exports.Foo.propTypes = { b: PropTypes.number };
```

After applying this commit, the "getMemberExpressionValuePath" utility
method will be able to find members defined on the same object as the
component definition that has been passed in or tries to find the member
defined on "(module.)exports" but does not check (!) if the passed-in
component definition is also being exported / assigned to module.exports

### DIFF
--- a/src/utils/__tests__/getMemberExpressionValuePath-test.js
+++ b/src/utils/__tests__/getMemberExpressionValuePath-test.js
@@ -53,6 +53,48 @@ describe('getMemberExpressionValuePath', () => {
         .toBe(def.parent.get('body', 1, 'expression', 'right'));
     });
 
+    it('finds nested property definitions', () => {
+      var def = statement(`
+        module.exports.Foo = () => {};
+        module.exports.Foo.propTypes = {};
+      `);
+
+      expect(getMemberExpressionValuePath(def, 'propTypes'))
+        .toBe(def.parent.get('body', 1, 'expression', 'right'));
+    });
+
+    it('finds the correct nested property definitions', () => {
+      var def = statement(`
+        module.exports.Foo = () => {};
+        module.exports.Foo.propTypes = {};
+        exports.Foo.propTypes = {};
+      `);
+
+      expect(getMemberExpressionValuePath(def, 'propTypes'))
+        .toBe(def.parent.get('body', 1, 'expression', 'right'));
+    });
+
+    it('finds property definitions that are part of a commonJS export', () => {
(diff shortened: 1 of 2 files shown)


src/utils/__tests__/getMemberExpressionValuePath-test.js
```diff
@@ -53,6 +53,48 @@ describe('getMemberExpressionValuePath', () => {
...
+      `);
+
+      expect(getMemberExpressionValuePath(def, 'propTypes'))
+        .toBe(def.parent.get('body', 1, 'expression', 'right'));
+    });
+
+    it('finds property definitions that are part of a commonJS export', () => {
+      var def = statement(`
+        function Foo () {}
+        module.exports.Foo.propTypes = {};
```
fkling: Overall I like it, but this, which I guess is what you were referring to, worries me a bit. This almost seems like too much magic (no pun intended wrt to your user name ;) )here.

In general I think the handler should be able to detect these cases:

```js
// Assignment to variable
var foo = <component definition>;
foo.propTypes = ...;

// Assignment to property
foo.bar = <component definition>;
foo.bar.propTypes = ...;

// In both cases we should be able to detect aliasing
var bar = foo; // = foo.bar;
bar.propTypes = ...;

something.else = foo; // = foo.bar;
something.else.propTypes = ...;
```

This is going to be a bit more elaborate I think. It basically requires us to 

- remember in which scope a component was defined
- look for assignments of the variable or member expression to another variable / member expression in that scope
- look for assignments to member expressions where the object is either the original variable / member expression or one of its aliases.

fkling:
Any thoughts about my comment @ZauberNerd ?

ZauberNerd:
Hey @fkling, yes and thank you for the comment, that's really valuable feedback and I agree, that the proposed solution is a  bit error prone. This is not a pressing issue for me at the moment, so I put the other issues before it, but I'll try to look into this again when I have some more time.

PR #114, opened 2016-08-09
Add resolver "findAllExportedComponentDefinitions"

ZauberNerd:
This resolver is basically a direct copy of the
"findExportedComponentDefinition" resolver including the tests.
The most significant difference is, that this resolver does not throw
when it encounters more than one component definition.
It adds all found component definitions into an array (deduplicated)
and returns that array at the end of execution.

### DIFF
--- a/src/resolver/findAllExportedComponentDefinitions.js
+++ b/src/resolver/findAllExportedComponentDefinitions.js
@@ -1,0 +1,121 @@
+/*
+ * Copyright (c) 2015, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @flow
+ */
+
+import isExportsOrModuleAssignment from '../utils/isExportsOrModuleAssignment';
+import isReactComponentClass from '../utils/isReactComponentClass';
+import isReactCreateClassCall from '../utils/isReactCreateClassCall';
+import isStatelessComponent from '../utils/isStatelessComponent';
+import normalizeClassDefinition from '../utils/normalizeClassDefinition';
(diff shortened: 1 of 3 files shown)


fkling:
This is great, thank you! I was just thinking about this the other day.

ZauberNerd:
Ah, great. Glad to see things go into upstream :)
I'm currently re-thinking my approach to a living style guide generator for react components (https://github.com/nerdlabs/schaufenster) and I think I'll just go with react-docgen and a bunch of custom handlers instead. I started writing a couple of them already [here for file-based dependencies](https://github.com/ZauberNerd/react-docgen/commit/6bc7e1c224ca4d080f4554bd730c809338c948c3) and [here for the export of the component](https://github.com/ZauberNerd/react-docgen/commit/120ac05fe9c422410912757b9f88c4f49b5f1229) but those are not really related to this project, so I didn't create PRs for them but will extract them into their own npm modules to allow others to use them too.

fkling:
> I didn't create PRs for them but will extract them into their own npm modules to allow others to use them too.

That's perfect. I will actually post something very soon about how I see react-docgen moving forward, and that's exactly what I had in mind!

fkling:
Could you export this handler from `src/resolver/index.js` as well?

src/resolver/findAllExportedComponentDefinitions.js
```diff
@@ -0,0 +1,121 @@
...
+ *
+ * modules.exports = Definition;
+ * exports.foo = Definition;
+ * export default Definition;
+ * export var Definition = ...;
+ */
+export default function findExportedComponentDefinitions(
+  ast: ASTNode,
+  recast: Object
+): ?NodePath {
```
fkling: Seems like you need to change this type annotation to `Array<NodePath>`.

ZauberNerd: I tried changing it yesterday and got different errors. I'll look into it when I have a bit more time.

PR #118, opened 2016-08-18
Fix intersection and union flow types

danez:
This is basically #97 but without throwing an error.

I also enabled to eslint rules to detect undefined and unused variables.

Closes #97

### DIFF
--- a/src/handlers/__tests__/flowTypeDocBlockHandler-test.js
+++ b/src/handlers/__tests__/flowTypeDocBlockHandler-test.js
@@ -127,6 +127,52 @@ describe('flowTypeDocBlockHandler', () => {
       .not.toThrow();
   });
 
+  it('supports intersection proptypes', () => {
+    var definition = statement(`
+      (props: Props) => <div />;
+
+      var React = require('React');
+      import type Imported from 'something';
+
+      type BarProps = {
+        /** bar description */
+        bar: 'barValue'
+      }
+
+      type Props = Imported & BarProps & {
+        /** foo description */
+        foo: 'fooValue'
+      };
+    `).get('expression');
+
+    flowTypeDocBlockHandler(documentation, definition);
+
+    expect(documentation.descriptors).toEqual({
+      foo: {
+        description: 'foo description',
+      },
+      bar: {
+        description: 'bar description',
+      },
+    });
+
+    it('does not support union proptypes', () => {
+      var definition = statement(`
+        (props: Props) => <div />;
+
+        var React = require('React');
+        import type Imported from 'something';
+
+        type Other = { bar: 'barValue' };
+        type Props = Imported | Other | { foo: 'fooValue' };
+      `).get('expression');
+
+      expect(() => flowTypeDocBlockHandler(documentation, definition))
+        .toThrowError(TypeError, "react-docgen doesn't support Props of union types");
+    });
+  });
+
+
   describe('does not error for unreachable type', () => {
     function test(code) {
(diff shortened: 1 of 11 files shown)


src/handlers/__tests__/flowTypeDocBlockHandler-test.js
```diff
@@ -127,6 +127,52 @@ describe('flowTypeDocBlockHandler', () => {
...
+
+        var React = require('React');
+        import type Imported from 'something';
+
+        type Other = { bar: 'barValue' };
+        type Props = Imported | Other | { foo: 'fooValue' };
+      `).get('expression');
+
+      expect(() => flowTypeDocBlockHandler(documentation, definition))
+        .toThrowError(TypeError, "react-docgen doesn't support Props of union types");
```
fkling: Shouldn't this *not* throw as well?

src/utils/getFlowTypeFromReactComponent.js
```diff
@@ -51,7 +50,48 @@ export default (path: NodePath): ?NodePath  => {
...
+    path.get('properties').each(
+      propertyPath => callback(propertyPath)
+    );
+  } else if (path.node.type === 'IntersectionTypeAnnotation') {
+    path.get('types').each(
+      typesPath => applyToFlowTypeProperties(typesPath, callback)
+    );
+  } else if (path.node.type !== 'UnionTypeAnnotation') {
+    // The react-docgen output format does not currently allow
+    // for the expression of union types
```
fkling: What would we have to do to make this work?

fkling:
Thanks for that! I added two comments which I guess don't show in here because the branch got deleted. But they should be visible when you click on "Files".

src/handlers/__tests__/flowTypeDocBlockHandler-test.js
danez: good catch. The test wasn't executed because of wrong nesting. Fixed it.

src/utils/getFlowTypeFromReactComponent.js
danez: This depends on how exactly we want to go for this. Quoting the flow docs:

> A union type requires for a value to be one of the input types.
> 
> ```js
> /* @flow */
> type U = number | string;
> var x: U = 1;
> x = "two";
> ```
> An intersection type requires a value to be all of the input types:
> 
> ```js
> /* @flow */
> type I = {a: number} & {b: number};
> var x: I = {a: 1, b: 2};
> x = {a: 1, b: 2, c: "three"};
> ```

So intersection was quite straight forward, just merge all objects and the result is the object with all possible props.

With union this is more complex. In [this](https://flowtype.org/try/#0PQKgBAAgZgNg9gdzCYAoVYAuBPADgUzAFUwBeMAbwEMAuMAOwFcBbAI3wCcBfMAH0tZ0Azpg4BLegHMuAbgwA3KhzAAPOiXLU6ARlkYVZSrTAAiEeKkm9qwxWsHNxk1SEmANGEGmXAEytygA) example, one can see that when using union for prop-types it is one of the objects to be "unioned" (with additional properties allowed).

So in the simple case of 
```
 type U = {a: number} | {a: string, b: string};
```
The resulting docs would need to be: "prop a is required and needs to be a number, or it is a string but then also b is required and needs to be a string"

This might make sense in some react components, but I'm not sure yet how we would document this cases. We maybe could only add the raw value.

P.S.: I might be mistaken here, but I think it should be correct. Never used union and intersect in these cases myself. :)

Issue #65, opened 2016-03-24
variables for propTypes


khankuan:
Hi, im using `oneOf(myArray)`. The problem is because myArray is a variable and declared above the component.

fkling:
Can you elaborate? What exactly is the problem? Do you get an error? Do you get unexpected output? What is the output your get and what is the output you would want to get?

khankuan:
sample snippet
```
const types = ['dark', 'light'];
...
static propTypes = {
  types: React.PropTypes.oneOf(types)
}

```
This resulted in the value being 'types' as a string. It would be nice if we can replace it with the variable value instead of the variable name.

ondrejbartas:
👍

fkling:
This should be fixed with d3dbb2e0214c8d5e84089c25a970479186e95383 which is included in the v2.12.0 release.

leepowellcouk:
Sorry to raise an old issue, but should this work if the values are kept in an imported file?

```
import { types } from './types';
...
static propTypes = {
  types: React.PropTypes.oneOf(types)
}
```

fkling:
@leepowellcouk: No, react-docgen doesn't resolve dependencies.

fchwpo:
Its not working for oneOfType

fchwpo:
Like in my case 
`const columnProps = [
	PropTypes.bool, //can be a boolean
	PropTypes.number, //can be a number
	PropTypes.string, //can be a string,
	PropTypes.shape({
		size: PropTypes.oneOfType([PropTypes.bool, PropTypes.number, PropTypes.string]), // An object that could be one of many types
		order: stringOrNumber,
		offset: stringOrNumber
	})
];

const propTypes = {
	/**
	 * The number of columns to span on extra small devices (<576px) 
	 * 
	 */
	xs : PropTypes.oneOfType(columnProps),
	/**
	 * The number of columns to span on small devices (≥576px)
	 * 
	 */
	sm : PropTypes.oneOfType(columnProps),
	/**
	 * The number of columns to span on medium devices (≥768px)
	 * 
	 */
	md : PropTypes.oneOfType(columnProps),
	/**
	 * The number of columns to span on large devices (≥992px)
	 * 
	 */
	lg : PropTypes.oneOfType(columnProps),
	/**
	 * The number of columns to span on extra large devices (≥1200px)
	 * 
	 */
	xl : PropTypes.oneOfType(columnProps),
	/**
	 * Custom ClassName String
	 */
[...]

PR #122, opened 2016-09-05
Resolve values in "oneOf" PropTypes

ZauberNerd:
This will allow the propTypesHandler generate better results for "oneOf" propTypes.
This change is made in two commits, the first one will enable the `resolveToValue` method
to resolve `MeberExpression`s to their initialization value and the second one will
use that to find the value of all Identifiers inside a `oneOf` propType.

Before:
```javascript
var TYPES = ["FOO", "BAR"];
PropTypes.oneOf(TYPES);
```
would print this as "custom".
Also
```javascript
var TYPES = { FOO: "FOO", BAR: "BAR" };
PropTypes.oneOf([TYPES.FOO, TYPES.BAR]);
```
would resolve to two values, but both are reported as computed custom values.

After:
Now the propTypesHandler will be able to resolve most of the common patterns that resolve to
static values inside the same module.

### DIFF
--- a/src/utils/__tests__/getPropType-test.js
+++ b/src/utils/__tests__/getPropType-test.js
@@ -122,20 +122,124 @@ describe('getPropType', () => {
     });
   });
 
-  it('resolves variables to their values', () => {
-    var propTypeExpression = statement(`
-      PropTypes.shape(shape);
-      var shape = {bar: PropTypes.string};
-    `).get('expression');
+  describe('resolve identifier to their values', () => {
+    it('resolves variables to their values', () => {
+      var propTypeExpression = statement(`
+        PropTypes.shape(shape);
+        var shape = {bar: PropTypes.string};
+      `).get('expression');
 
-    expect(getPropType(propTypeExpression)).toEqual({
-      name: 'shape',
-      value: {
-        bar: {
-          name: 'string',
-          required: false,
+      expect(getPropType(propTypeExpression)).toEqual({
+        name: 'shape',
+        value: {
+          bar: {
+            name: 'string',
+            required: false,
+          },
         },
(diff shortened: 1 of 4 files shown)


fkling:
This is awesome!

I'm a little bit worried about the fact that `resolveToValue` used to simply return the MemberExpression if you passed one and now it tries to resolve it. While that behavior is certainly better, it changed.

Since we might update the parser to flow anyway, maybe well just make this part of a breaking change release.

This would fix #65 btw.

ZauberNerd:
Yes I think `resolveToValue` has been named a bit misleading and maybe there should be two methods: `resolveToInitialization` (or declaration) and `resolveToValue`.
The former would traverse the scope chain and find module imports or other variable declarations, while the latter would try to find the actual value ((re)assignment of a literal value, which might involve traversing through function return values or aliasing). That would of course not always be possible or maybe not very correct, but could be helpful in some cases.

Using flow as parser sounds very interesting, is this being done in the open and has work on it already started?

fkling:
> Using flow as parser sounds very interesting, is this being done in the open and has work on it already started?

@wbinnssmith started to look into this. In the best case, we literally can just change [the line](https://github.com/reactjs/react-docgen/blob/master/src/parse.js#L53)

```js
var ast = recast.parse(src, {esprima: babylon});
```

to

```js
var ast = recast.parse(src, {esprima: flow});
```

and everything still works the same.

fkling:
I will resolve the conflict and merge this. I don't think there will be any issues that `resolveToValue` changed slightly.

fkling:
Merged with d3dbb2e0214c8d5e84089c25a970479186e95383 .

Issue #80, opened 2016-05-25
Redux connect support


maximelebastard:
Hi,

I have a React/Redux app that I'd like to document. With redux, I use the [connect()](https://github.com/reactjs/react-redux/blob/master/docs/api.md) function.

But when I use it, react-docgen cannot find a valid component in my file.

This works:

````
const Contributors = React.createClass({...});
export default Contributors;
````

This does not:

```
const Contributors = React.createClass({...});
const mapStateToProps = (state) => {
  return { contributors: state.contributors }
}
export default connect(mapStateToProps)(Contributors);
```

I found a workaround by appending this to the end of file, but I don't know if it will not break something later:

```
module.exports = Contributors;
```

**Comment by rtsao** [comment]
I've run into this issue as well. I've added support for single unary higher-order-components in my fork. This is the relevant commit here: https://github.com/rtsao/react-docgen/commit/46ec8dfedc323ff1608229ae8d1c3064669455d4

You can turn your code to use a single unary higher-order component like so:

```js
const Contributors = React.createClass({...});
const mapStateToProps = (state) => {
  return { contributors: state.contributors }
}

const hoc = (Component) => connect(mapStateToProps)(Component);

export default hoc(Contributors);
```

The above code will work with my fork.

If multiple higher order components are needed, you can use [recompose](https://github.com/acdlite/recompose)'s `compose` helper.

I think **only** supporting a single unary higher-order component seems like a reasonable compromise, since basically you can convert all forms of higher-order component composition into that form. Supporting multiple call expressions seems like you could quickly go down a rabbit hole.

I'd be happy to create a PR if @fkling thinks my approach is reasonable.

fkling:
@rtsao:

> I've added support for single unary higher-order-components in my personal fork

That seems very reasonable to me, and I think it would even make sense to expand this to non-unary functions. The other arguments could be included in the output as well, as far as possible. Happy to review that PR :)

@maximelebastard
As an intermediate solution, if you can use the `findAllComponentDefinitions` resolver, which looks for all component definitions, not just the exported ones.

tizmagik:
@rtsao that looks very useful! Was a PR ever submitted for this?

rtsao:
Yeah, actually I think this should be very doable.

```js
export default hoc1(Component);

export default hoc1(arg1a, arg1b)(Component);

export default hoc2(arg2b, arg2b)(
  hoc1(arg1a, arg2a)(Component)
);

[...]

PR #124, opened 2016-09-14
Added support for resolving exported components within HOCs

rtsao:
Fixes https://github.com/reactjs/react-docgen/issues/80

### DIFF
--- a/src/utils/resolveHOC.js
+++ b/src/utils/resolveHOC.js
@@ -1,0 +1,38 @@
+/*
+ * Copyright (c) 2015, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @flow
+ *
+ */
+
+import recast from 'recast';
+import isReactCreateClassCall from './isReactCreateClassCall';
+
+var {
+  types: {
+    NodePath,
+    namedTypes: types,
+  },
+} = recast;
+
+/**
+ * If the path is a call expression, it recursively resolves to the
+ * rightmost argument, stopping if it finds a React.createClass call expression
+ *
+ * Else the path itself is returned.
+ */
+export default function resolveHOC(path: NodePath): NodePath {
+  var node = path.node;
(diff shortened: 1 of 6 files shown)


rtsao:
FWIW I've signed the CLA (as seen on https://github.com/reactjs/react-docgen/pull/71) but I'm not sure why it isn't show up here

fkling:
> but I'm not sure why it isn't show up here

The reactjs org is not managed by Facebook anymore afaik.

src/utils/resolveHOC.js
```diff
@@ -0,0 +1,38 @@
...
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @flow
+ *
+ */
+
+import recast from 'recast';
+import isReactCreateClassCall from './isReactCreateClassCall';
```
ZauberNerd: What about `isReactComponentClass` and `isStatelessComponent`?

src/utils/resolveHOC.js
```diff
@@ -0,0 +1,38 @@
...
+var {
+  types: {
+    NodePath,
+    namedTypes: types,
+  },
+} = recast;
+
+/**
+ * If the path is a call expression, it recursively resolves to the
+ * rightmost argument, stopping if it finds a React.createClass call expression
```
ZauberNerd: Inspecting only the rightmost argument may not work for all cases.
One common case is for example Relay, where the component will be the [first argument](https://facebook.github.io/relay/docs/guides-containers.html#relay-containers).

Also it might be interesting to investigate `connect()` of [react-redux](https://github.com/reactjs/react-redux/blob/master/docs/api.md) and [recompose](https://github.com/acdlite/recompose/blob/master/docs/API.md), as a baseline for example usage of HoCs.

rtsao: That's interesting that Relay doesn't follow the convention I've seen for every other HOC, which is to take only one argument (the component) or have the component be the rightmost argument. Recompose follows this convention.

But as discussed in https://github.com/reactjs/react-docgen/issues/80 you can easily turn any HOC into a HOC that takes a component as its only or rightmost argument.

fkling:
Yeah, looks good to me, I think we can always iterate on it and improve it if necessary. Would be cool to document this behavior though.

PR #127, opened 2016-09-27
fix(bin): write into streams before exit

mmeineke:
I run into an issue, that the output from `react-docgen` is cut off, if the output is too big.
This only happens if I try to pipe the output like so:
```
react-docgen ./components | node ./buildDocs.js
```
Or like so:
```
let spawn = require('child_process').spawn;
let child = spawn('react-docgen', ['./Components']);
let buffers = [];
child.stdout.on('data', data => {
    buffers.push(data);
});
child.stdout.on('close', () => {
    let buffer = Buffer.concat(buffers);
    console.log(`${buffer}`);
});
```
After some research I found this issue:
https://github.com/visionmedia/mocha/issues/333
And it works with the solution they used:
https://github.com/mochajs/mocha/pull/1440/files

So hopefully you can use this fix. I will try to help where I can to get it merged. Another solution could be this package: https://www.npmjs.com/package/exit

### DIFF
--- a/bin/react-docgen.js
+++ b/bin/react-docgen.js
@@ -9,8 +9,6 @@
  *
  */
 
-/*eslint no-process-exit: 0*/
-
 var argv = require('nomnom')
   .script('react-docgen')
   .help(
@@ -90,10 +88,11 @@ if (argv.resolver) {
       if (e.code !== 'MODULE_NOT_FOUND') {
         throw e;
       }
-      exitWithError(
+      writeError(
         `Unknown resolver: "${argv.resolver}" is neither a built-in resolver ` +
         `nor can it be found locally ("${resolverPath}")`
       );
+      process.exitCode = 1;
     }
   }
 }
@@ -112,12 +111,7 @@ function writeError(msg, filePath) {
   }
 }
 
-function exitWithError(error) {
-  writeError(error);
-  exit(1);
-}
-
-function exitWithResult(result) {
+function writeResult(result) {
   result = argv.pretty ?
     JSON.stringify(result, null, 2) :
     JSON.stringify(result);
@@ -126,7 +120,6 @@ function exitWithResult(result) {
   } else {
     process.stdout.write(result + '\n');
   }
-  exit(0);
 }
 
 function traverseDir(filePath, result, done) {
@@ -139,7 +132,7 @@ function traverseDir(filePath, result, done) {
     },
     function(error, content, filename, next) {
       if (error) {
-        exitWithError(error);
+        throw error;
       }
       try {
         result[filename] = parse(content);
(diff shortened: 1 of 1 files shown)


danez:
With which node version do you have the problem? As far as I see it was fixed in node 0.12.

mmeineke:
I tried it with node 4 and 6. Was wondering as well why I ran into this issue

mmeineke:
I will provide you with an example

mmeineke:
I added 2 examples which uses exactly the same code as in the examples in this repo beside the amount of props. The first example with a component with 100 props and the second with 1000 props. Please try if the second one fails for you as well.
https://github.com/mmeineke/react-docgen-test

danez:
Thanks, yes I can reproduce this.

Reading more about writing to stdout I found this in the docs:
> The process.stderr and process.stdout streams are blocking when outputting to TTYs (terminals) on OS X as a workaround for the operating system's small, 1kb buffer size. This is to prevent interleaving between stdout and stderr.
> 
> To check if Node.js is being run in a TTY context, check the isTTY property on process.stderr, process.stdout, or process.stdin.
> 
> For instance:
> 
> $ node -p "Boolean(process.stdin.isTTY)"
> true
> $ echo "foo" | node -p "Boolean(process.stdin.isTTY)"
> false

So as you are piping we are not in a tty anymore which makes the stream write non-blocking and the `process.exit()` immediately afterwards might interrupt the writing.

Instead of doing this draining logic, I would more lean to remove `process.exit()` from line 98 and 110 and replace the first one with `process.exitCode = 1;`. The process is ending after the output anyway, I don't see a need to manually do the exit call.

@fkling What do you think?

fkling:
> Instead of doing this draining logic, I would more lean to remove process.exit() from line 98 and 110 and replace the first one with `process.exitCode = 1;`.

If that works, I'm all for it since it's the simpler solution. However, we are invoking `exitWithError` in  line 142 which is the async directory traversal process. Maybe we can just `return` there instead?

mmeineke:
Hey, I removed the process.exit completely now.
- renamed `exitWithResult` to `writeResult`
- removed `exitWithError` (added `process.exitCode = 1;` instead)
- `traverseDir` throws an error now which gets catched

What I don't like and hopefully you have a better idea is line `153` `if (process.exitCode !== 1) {` which I needed to implement because of the removed `exitWithResult` in line `91/95`

mmeineke:
@danez , @fkling what do you think?

bin/react-docgen.js
```diff
@@ -90,10 +88,11 @@ if (argv.resolver) {
       if (e.code !== 'MODULE_NOT_FOUND') {
         throw e;
       }
-      exitWithError(
+      writeError(
         `Unknown resolver: "${argv.resolver}" is neither a built-in resolver ` +
         `nor can it be found locally ("${resolverPath}")`
       );
+      process.exitCode = 1;
```
fkling: Couldn't we just call `process.exit(1)` here?

mmeineke:
I replaced the `writeError`+ `process.exit` with an `errorMessage` which could be used for other error messages in the future as well.
This errorMessage gets written below with the `process.exit`.
Would be fine for me now.

mmeineke:
@danez Hey, could we merge the PR now?

mmeineke:
@danez thanks a lot :)

fkling:
Thanks for this! I'll release a new version shortly.

Issue #131, opened 2016-10-21
Incorrect default value inferring for stateless components


arunoda:
This issue came with the version `2.12.0`.
And we got this while working with our [babel plugin for doc gen](https://github.com/kadirahq/babel-plugin-react-docgen).

Assume we've a stateless component like this:

```js
const Button = ({ children, onClick, style = {} }) => (
  <button
    style={{ }}
    onClick={onClick}
  >
    {children}
  </button>
);

Button.propTypes = {
  children: React.PropTypes.string.isRequired,
  onClick: React.PropTypes.func,
  style: React.PropTypes.object,
};

export default Button;
```

Then generated info like this:

```js
{
  description: '',
  props: {
    children: {
      type: {
        name: 'string'
      },
      required: true,
      description: '',
      defaultValue: {
        value: 'children',
        computed: true
      }
    },
    onClick: {
      type: {
        name: 'func'
      },
      required: false,
      description: '',
      defaultValue: {
        value: 'onClick',
        computed: true
      }
    },
    style: {
      type: {
        name: 'object'
      },
      required: false,
      description: '',
      defaultValue: {
        value: '{}',
        computed: false
      }
    }
  }
}
```

Just look at the default values for `onClick` or `children`.
This doesn't seems to be correct to me.

fkling:
Thanks, I'll look into it!

fkling:
This should be fixed with v2.12.1!

arunoda:
This is pretty fast guys. Awesome.

PR #132, opened 2016-10-21
Fix issue with computed props for functional components

CompuIves:
Fixes issue https://github.com/reactjs/react-docgen/issues/131.

The issue was that props that didn't have any defaults were still evaluated (since the prop was indeed referenced) and received `{ value: propName, computed: 'true' }` as defaultProp. I also found an issue that imported values were not properly recognised by defaultprops for stateless components.

Both issues are fixed with this PR.
I can explain the issue more explicitly next week, I am gone this weekend.


fkling:
Sweet, thank you for the quick fix!

PR #134, opened 2016-11-08
Fixed infinite loop in utils/isStatelessComponent.js

rtsao:
Parsing a file that contains anything like below will result in react-docgen hanging:

```jsx
export function foo() {
  return [].join();
}
```

I haven't had a chance to dig into this yet, but we've been bitten by this.

### DIFF
--- a/src/__tests__/fixtures/component_2.js
+++ b/src/__tests__/fixtures/component_2.js
@@ -21,3 +21,7 @@ export class Button extends React.Component {
     };
   }
 }
+
+export function foo() {
+  return [].join();
+}
--- a/src/utils/isStatelessComponent.js
+++ b/src/utils/isStatelessComponent.js
@@ -94,6 +94,9 @@ function returnsJSXElementOrReactCreateElementCall(path) {
           else {
             while (calleeValue.get('object').node.type !== 'Identifier') {
               calleeValue = calleeValue.get('object');
+              if (calleeValue.node.type === 'ArrayExpression') {
+                break;
+              }
               namesToResolve.unshift(calleeValue.get('property'));
             }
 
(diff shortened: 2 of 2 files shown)


rtsao:
The infinite loop occurs in this while expression: https://github.com/reactjs/react-docgen/blob/6f7368658ddda6f7cdb03d62c0a59b065bdd42fe/src/utils/isStatelessComponent.js#L95

I'll see if I can fix this

rtsao:
I've added a really basic fix. I think array literal usage is the corner case here so a simple check should suffice. It doesn't seem to me a loop would occur in any other case.

fkling:
Uh, thanks for that! I believe there are other cases whether this would be an issue, e.g.

```js
return {}.foo()
```

It seems like what this is doing is unrolling bested *MemberExpression*s. So maybe the loop condition should rather be

```js
while (calleeValue.get('object').node.type === 'MemberExpression')
```

?

cc @iamdustan who wrote this code. Maybe adding a comment about what exactly this does would be useful 😉

iamdustan:
Hey, this was a long time ago :innocent:

I was out of town last week and only have a couple days around this week and next week; so it’ll probably be 3 weeks before I will have time to get my brain back into this for realsies.

fkling:
@rtsao, what's the progress or your thoughts on this?

rtsao:
I've been using my fork with this patch without any issue (I think call expressions with object literals is pretty rare).

I'd be happy to refactor this to make the fix more robust (maybe explore the alternate while condition described above) but probably won't get a chance to in the near term.

iamdustan:
After looking through both of these a bit more I am pretty sure that PR #139 that @tchon just opened is the more robust fix to this.

rtsao:
Agreed, closing in favor of https://github.com/reactjs/react-docgen/pull/139

PR #135, opened 2016-11-09
Add __mocks__ folder to default ignoreDir

aitherios:
As __mocks__ is used in manual mocks on jest
cf. https://facebook.github.io/jest/docs/manual-mocks.html


fkling:
Seems reasonable, thank you!

Issue #125, opened 2016-09-21
Flow polymorphic types cannot be parsed


aaronjensen:
I had this:

```js
export type Props = {
  handleSubmit: () => void,
  onBack?: () => void,
  children?: React.Element<*>,
  backText?: string,
  nextText?: string,
}
```

And it failed to parse so I had to change my children type to `mixed`. I'm guessing this has to do w/ the fact that docgen is using an old version of babylon.

fkling:
Yeah, I think we should simply use the flow parser since it's more ESTree-compatible than babylon 6.

aaronjensen:
babylon 6 can handle all of these though, right? Is it just a version issue?

The flow parser is ocaml, so I'm not sure it could be used here.

fkling:
There is JS version of the flow parser available: https://www.npmjs.com/package/flow-parser . I'd rather use flow since I know they are trying to stay close to the ESTree spec.

PR #137, opened 2016-11-18
Use Flow's parser instead of babylon

wbinnssmith:
This switches react-docgen's parser to the Flow parser, ensuring better
compatibility with the JavaScript flow can parse. react-docgen was
previously limited to whatever Babel 5 could parse.

Current caveats:
 * Flow's parser currently [incorrectly parses decorators](https://github.com/facebook/flow/issues/2741),
   leading to a host of workarounds in the jsdoc docblock-finding code.
   Once the flow parser can correctly consider these decorators and
   include the "@" character in the decorator source rather than a free
   identifier, recast should be able to correctly attribute the docblock
   comment, letting us remove most, if not all of these hacks.
 * Also due to Flow's parsing of decorators as two identifiers, using
   a decorated class expression as an expression (valid JavaScript, see
   below) fails. This is a breaking change with current react-docgen
   behavior. We'll need to set the baseline for docgen as whatever Flow
   can correctly parse. Here's an example of valid JS that breaks in
   Flow:

   ```js
   myFunction(
     /**
     * My Docblock
     */
     @MyDecorator
     class MyClass extends React.Component {}
   )
   ```

Thanks for your help on this personally, @fkling!
Hopefully this opens up a whole new set of users of docgen :)


fkling:
Thank you! This will fix #125 .

wbinnssmith:
cc @mroch @gabelevi on the Flow issues, not sure how much a priority (arguably deprecated) decorators are, but I imagine that failing test case might help with facebook/flow#2741

danez:
But with this we will loose the ability to parse early stage proposals, which babylon supports, so babylon 6 could be a valid alternative here.

fkling:
Good point @danez, I haven't thought about that. However, babylon 6 deviates from the estree spec which also isn't great.

I'd be curious how much the early proposals are actually used.

aaronjensen:
I wonder if going the route of eslint and having pluggable parsers is the right way. It's too bad we can't just have one 😄

MoOx:
I am having issue with react-styleguidist when trying to make my files using this kind of syntax: `name?: value`. babylon fails on ?. But not sure why since for my app I don't have any issue and babylon parse this correctly as babel transforms this like a charm.

danez:
@MoOx react-docgen is using babylon 5 which does not support newer flow features.

MoOx:
Oh! Make sense then. Thanks for the comment :)

danez:
What we were also discussing some time ago in the babel slack, if it would make sense to have an option to create estree compatible output (at least for stage-4 syntax). This could then at least help the effort to have exchangeable parsers in react-docgen and one of them being babylon6.

fkling:
@danez Any update on that? I started looking into supporting both, babylon and flow, but it's a bit of work. Having babylon produce an estree compliant AST would certainly make this a lot easier 😉

danez:
Yes I started on that already, but will need more time babel/babylon#277

fkling:
@danez This looks certainly better than what I currently have to do to make react-docgen work. Thank you!

jribeiro:
@danez I see the your feature babel/babylon#277 was merged but this PR is still opened. Do you know if this is this still an issue?

danez:
Yes you can try install react-docgen v3.0.0-beta5 which includes the latest babylon version.

jribeiro:
Great @danez, will do. Thanks a lot

danez:
I close that for now, we can get back to this if we want to have exchangeable parsers.

Issue #136, opened 2016-11-10
Hangs forever on chained functions


garamond:
Input file `MyComp.js`:
```
export class MyComp extends Component { render(){} }

export function myFun(foo) {
  return foo.bar().baz()
}
```

Running `react-docgen MyComp.js` hangs forever and causes CPU load to spike to 100%.

Without chained functions the issue does not appear: After changing to body of `myFun` to
```
export function myFun(foo) {
  return foo.bar.baz()
}
```
the file is processed without issues.

fkling:
Thanks! This will be resolved with #134 .

tchon:
@fkling It looks like https://github.com/reactjs/react-docgen/pull/134 will not address the hanging on chained methods. See improved PR here: https://github.com/reactjs/react-docgen/pull/139

PR #139, opened 2016-11-29
Guard against infinite loop on chained or static methods

tchon:
### Summary
  * Fixes https://github.com/reactjs/react-docgen/issues/136
  * Add `isValidCalleeType()` check for CallExpresion and ArrayExpression and Literal
  * Improves on https://github.com/reactjs/react-docgen/pull/134 by validating the `calleeValue`'s object type, which happens to be mutated in-place during the `while` loop.

### DIFF
--- a/src/__tests__/fixtures/component_2.js
+++ b/src/__tests__/fixtures/component_2.js
@@ -21,3 +21,11 @@ export class Button extends React.Component {
     };
   }
 }
+
+export function foo() {
+  return [].join();
+}
+
+export function chained() {
+  return foo.bar().join();
+}
--- a/src/__tests__/fixtures/component_2.js
+++ b/src/__tests__/fixtures/component_2.js
@@ -29,3 +29,7 @@ export function foo() {
 export function chained() {
   return foo.bar().join();
 }
+
+export function templateLiteral() {
+  return `foo bar`.split(' ');
+}
--- a/src/utils/isStatelessComponent.js
+++ b/src/utils/isStatelessComponent.js
@@ -27,6 +27,15 @@ const validPossibleStatelessComponentTypes = [
   'ArrowFunctionExpression',
 ];
 
+function isValidCalleeType(type) {
+  return [
+    'Identifier',
+    'CallExpression',
+    'ArrayExpression',
+    'Literal'
+  ].indexOf(type) < 0;
+}
+
 function isJSXElementOrReactCreateElement(path) {
   return (
     path.node.type === 'JSXElement' ||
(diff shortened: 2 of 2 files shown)


flut1:
I recently had the hanging issue on a template literal, even after applying this PR. After adding the type `TemplateLiteral` to the list of types in `isValidCalleeType` the issue was resolved

iamdustan:
@flut1 could you include a code sample we could add to the test showing that?

tchon:
@flut1 I've added the `TemplateLiteral` to list of *invalid* callee types.  If you can provide a code example, then I can also add it to the spec fixtures. Thanks!

flut1:
@tchon 
```
export function stringLiteral() {
  return `foo bar`.split(' ');
}
```

rhalff:
Experiencing the same issue and glad to find there is already a pull request to solve this issue, however it does not seem to cover all cases.

A (mixin) declaration like this will also hang:
```js
export default function withThis() {
  return this.foo();
}
```

The contents of calleeValue within https://github.com/reactjs/react-docgen/blob/master/src/utils/isStatelessComponent.js#L96 will continuously be:
```js
NodePath {
  value: undefined,
  parentPath: 
   NodePath {
     value: undefined,
     parentPath: 
      NodePath {
        value: undefined,
        parentPath: [Object],
        name: 'object',
        __childCache: [Object] },
     name: 'object',
     __childCache: { property: [Circular] } },
  name: 'property',
  __childCache: null }
```

MoOx:
Maybe better than nothing. Cause this hanging are a pain to debug sometimes...

rhalff:
Agreed, in which case I'm just suggesting to also add the `'ThisExpression'` to the invalid types.

danez:
What if we just simply do:
```
@@ -102,7 +102,7 @@ function returnsJSXElementOrReactCreateElementCall(path) {
             resolvedValue = resolveToValue(calleeValue.get('object'));
           }
           else {
-            while (calleeValue.get('object').node.type !== 'Identifier') {
+            while (calleeValue.node.object && calleeValue.get('object').node.type !== 'Identifier') {
               calleeValue = calleeValue.get('object');
               namesToResolve.unshift(calleeValue.get('property'));
             }
```

All tests (including the ones from this PR) are all working, and not hanging.

danez:
I created a new PR with your commits, plus some changes that it exits the loop as soon as it hits a non-MemberExpression node. What do you think?

tchon:
👍 thanks @fkling @danez

fkling:
Sorry for the long wait!

PR #142, opened 2016-12-14
Replace nomnom with commanderjs

### LINKED ISSUE #56: Replace nomnom with commander.js
nomnom is not maintained anymore, so it might be a good idea to switch to an alternative.
https://github.com/harthur/nomnom

commander.js is one of them
https://github.com/tj/commander.js

withinboredom:
Fixes #56 

This is a pretty straight port over to commander.

Benefits:
- Removes dependency on a dead package.
- Commander only has one dependency: graceful-readlink, making install times just a tiny tiny bit faster ;)

### DIFF
--- a/bin/react-docgen.js
+++ b/bin/react-docgen.js
@@ -17,35 +17,36 @@ function collect(val, memo) {
 }
 
 argv
+    .usage('[path...] [options]')
+    .description(
+        'Extract meta information from React components.\n' +
+        '  If a directory is passed, it is recursively traversed.')
     .option(
-        '--path <PATH>',
-        'A component file or directory. If no path is provided, it reads from stdin.', collect, [])
-    .option(
-        '-o, --out <FILE>',
+        '-o, --out <file>',
         'Store extracted information in the FILE')
     .option(
         '--pretty',
         'pretty print JSON')
     .option(
-        '-x, --extension <EXTENSION>',
+        '-x, --extension <extension>',
         'File extensions to consider. Repeat to define multiple extensions. Default: js, jsx',
         collect,
         ['js', 'jsx'])
     .option(
-        '-e, --exclude <PATH>',
+        '-e, --exclude <path>',
         'Filename pattern to exclude.',
         collect,
         [])
     .option(
-        '-i, --ignore <PATH>',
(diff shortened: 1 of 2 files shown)


bin/react-docgen.js
```diff
@@ -9,55 +9,45 @@
  *
  */
 
-var argv = require('nomnom')
-  .script('react-docgen')
-  .help(
-    'Extract meta information from React components.\n' +
-    'If a directory is passed, it is recursively traversed.'
```
danez: This message is not there anymore and I think it is nice to have.

bin/react-docgen.js
```diff
@@ -9,55 +9,45 @@
...
+var argv = require('commander');
+
+function collect(val, memo) {
+  memo.push(val);
+  return memo;
+}
+
+argv
+    .option(
+        '--path <PATH>',
```
danez: path was not an option but an argument
from the help with nomnom:
```
Usage: react-docgen [path]... [options]
```

So this option would be new, and is also unused.
seems that `[FILES]` is the equivalent to `[path]`

bin/react-docgen.js
```diff
@@ -9,55 +9,45 @@
...
+    .option(
+        '-i, --ignore <PATH>',
+        'Folders to ignore',
+        collect,
+        ['node_modules', '__tests__', '__mocks__'])
+    .option(
+        '--resolver <RESOLVER>',
+        'Resolver name (findAllComponentDefinitions, findExportedComponentDefinition) or path to a module that exports a resolver.',
+        'findExportedComponentDefinition')
+    .arguments('[FILES]');
```
danez: As mentioned above this probable can be `[path]`

bin/react-docgen.js
withinboredom: Oh, my mistake. Will fix asap.

withinboredom:
@danez I've addressed those issues if you'd like to take another look

bin/react-docgen.js
withinboredom: Added back :)

danez:
Thank you so much for doing this :)

withinboredom:
No problem, glad to be of service 👍

PR #143, opened 2016-12-14
Fix endless loop

danez:
Follow up of #139


fkling:
Yeah, that seems the most reasonable thing to do. Listing every expression that is valid inside a MemberExpression seems cumbersome. I'll merge the PRs one by one to preserve contributions.

Thank you!

PR #145, opened 2016-12-16
Ignore SpreadProperties in shapes

orktes:
Prevent react-docgen from crashing when spread operations are passed to shape(). Behaviour is similar to how spread operations are handled in normal propType definitions.


fkling:
Thank you!

PR #146, opened 2016-12-21
[issue] Illustrate a syntax error with ES2016

oliviertassinari:
> 
      SyntaxError: Unexpected token (3:24)      
          at Parser.pp.raise (node_modules/babylon/lib/parser/location.js:24:13)
          at Parser.pp.unexpected (node_modules/babylon/lib/parser/util.js:82:8)
          at Parser.pp.parseExprAtom (node_modules/babylon/lib/parser/expression.js:425:12)

I'm not sure where to start to address that issue.
Here is the proposal for reference https://github.com/rwaldron/exponentiation-operator.

### DIFF
--- a/src/__tests__/fixtures/component_5.js
+++ b/src/__tests__/fixtures/component_5.js
@@ -1,5 +1,7 @@
 import React from 'react';
 
+const RANDOM_VALUE = 2 ** 2;
+
 const Button = ({ children, onClick, style = {} }) => (
   <button
     style={{ }}


danez:
I'm not sure when the exponential operator was introduced in babylon, but it might be that babylon 5 has not support for it. This is one more reason to decide for a solution on this: #137

oliviertassinari:
I guess that upgrading to Babylon 6 isn't straightforward. Thanks for the answer.

oliviertassinari:
I'm closing that PR, I don't need it open. I have made my point.

fkling:
Thank you @oliviertassinari! We are waiting until the estree plugin for babylon is ready, then we are going to upgrade to it with the option to use the flow parser as well.

PR #156, opened 2017-02-13
fix(docblocks): avoid breaking on expected Object prototype methods

### LINKED ISSUE #155: Any reason to use Object.create(null)?
We use react-docgen in [React Styleguidist](https://github.com/styleguidist/react-styleguidist) and we also use [to-ast](https://github.com/devongovett/to-ast) to generate JavaScript code in webpack loaders. to-ast don’t work with `doclets` generated by react-docgen because it tries to call `hasOwnProperty` and fails. I’m not really sure where should I send a pull request: to react-docgen or to to-ast, or I should make some work around in my code. In my understanding `Object.create(null)` is unnecessary here and just `{}` would work fine but I may miss something.

[This line](https://github.com/reactjs/react-docgen/blob/d65bfc460d7274206c3208fc41518727acd5a490/src/utils/docblock.js#L58) in react-docgen cause an error in [this line](https://github.com/devongovett/to-ast/blob/master/index.js#L128) of to-ast:

```
TypeError: obj.hasOwnProperty is not a function
    at Array.map (native)
```

AoDev:
Should fix #155

### DIFF
--- a/src/utils/docblock.js
+++ b/src/utils/docblock.js
@@ -55,7 +55,7 @@ export function getDocblock(path: NodePath): ?string {
  * and their "content" as values.
  */
 export function getDoclets(str: string): Object {
-  var doclets = Object.create(null);
+  var doclets = {};
   var match = DOCLET_PATTERN.exec(str);
 
   for (; match; match = DOCLET_PATTERN.exec(str)) {


fkling:
Thanks @AoDev! However, I think this is better solved in styleguidist by using something like `Object.prototype.hasOwnProperty.call(obj, ...)`.

AoDev:
It does not break in styleguidist but rather in a third party dependency on
which one may or may not have control.

I understand the situation, and that you prefer to have "clean" key value
store. So, more as a personal curiosity. What do you think would happen if
everybody started to create objects randomly this way? Wouldn't this break
a lot of code relying on this, although valid?

I understand the object is not "owned" by us, but then instead of checking
for hasOwnProperty now, should everyone starts checking if the object has a
prototype although it's perfectly valid. What do you think?

fkling:
> It does not break in styleguidist but rather in a third party dependency on
which one may or may not have control.

Yeah, I realized that now (see the issue).

> Wouldn't this break a lot of code relying on this, although valid?

Probably yes. I think this is a funny situation though: One could argue that the whole `hasOwnProperty` check is already a workaround for exactly the problem that `Object.create(null)` solves: Avoiding that properties "bleed through" the prototype chain. In a perfect world there would be no need to use `hasOwnProperty` in this situation.

However, the best solution, on our side, would probably be to use `Map` instead.

>  instead of checking for hasOwnProperty now, should everyone starts checking if the object has a prototype although it's perfectly valid.

No, I think fewer assumptions need to be made about objects. Ideally `hasOwnProperty` would be moved the `Object` or `Reflect`, but it should also be used sparingly. Looking at the prototype (chain) or making any assumption about the prototype chain defeats its purpose IMO (i.e. the consumer should only care about what properties an object has but not *how* it provides them).

AoDev:
I was reading the [Map docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) on the Mozilla site when I answered. Obviously it's a problem because of the language.

What brings my attention on their site is:
```
An Object has a prototype, so there are default keys in the map that could collide with your keys if you're not careful. This could be bypassed by using map = Object.create(null) since ES5, but was seldom done.
```

I think that my personal practice on this is will be to avoid using `Object.create(null)` mainly because of a legacy usage of the JS language, although valid, unless I use `Map`.

AoDev:
But, anyway, I think you can close this PR unless you decide to accept the change. Thank you for your answers : )

PR #158, opened 2017-02-18
Update babylon to 7

danez:
This update babylon to the latest beta of 7.0. This version is the first to include the estree plugin.
~The estree plugin will be also released with a minor 6.x version.~

It looks really good, nearly no changes necessary, only some tests according to raw value are failing.

TODO:

- [x] depends on #157 
- [x] raw values differ from babylon 5 (will fix upstream)
- [x] ~ranges not supported in babylon >= 6 (will investigate upstream)~

### DIFF
--- a/src/utils/__tests__/resolveToValue-test.js
+++ b/src/utils/__tests__/resolveToValue-test.js
@@ -49,7 +49,7 @@ describe('resolveToValue', () => {
     );
   });
 
-  it('handles SpreadProperties properly', () => {
+  it('handles SpreadPElements properly', () => {
     var path = parse([
       'var {foo: {bar}, ...baz} = bar;',
       'baz;',
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   "dependencies": {
     "async": "^2.1.4",
     "babel-runtime": "^6.9.2",
-    "babylon": "~5.8.3",
+    "babylon": "7.0.0-beta.3",
     "commander": "^2.9.0",
     "doctrine": "^2.0.0",
     "node-dir": "^0.1.10",
--- a/src/babylon.js
+++ b/src/babylon.js
@@ -13,21 +13,20 @@ var babylon = require('babylon');
 var options = {
   sourceType: 'module',
   strictMode: false,
-  locations: true,
-  ranges: true,
-  ecmaVersion: 7,
-  features: {
-    'es7.classProperties': true,
-    'es7.decorators': true,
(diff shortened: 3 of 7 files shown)


MoOx:
Amazing. Looking forward to get this!

danez:
We probably want to merge this with #137 and make the parser configurable (Maybe similar to prettier)? I think @fkling had already plans for this.

Can also work on this if we can agree how it should work.

//cc @wbinnssmith

fkling:
Awesome! Yes, making the parser configurable was the idea because flow doesn't support everything that babylon supports. However, we could probably also just merge this PR to finally get support for new flow types.

How does prettier let you specify parsers?

danez:
Prettier has this file: https://github.com/prettier/prettier/blob/master/src/parser.js
and in the options it can be specified which parser to use. So still limited to flow/babylon, but one can choose between these two. They probably did this, because babylon (still) does not understand all flow features.

I would wait with merging till the next 6.x release is out (today/tomorrow), which has the same estree functionality as this beta version. (https://github.com/babel/babylon/blob/changelog-6.16/CHANGELOG.md)

fkling:
Ah I see. Yeah, that's exactly what I had in mind (we do pretty much the same thing in jscodeshift).

Thank you for this!

fkling:
So, it's not coming in v6.16.0?

danez:
The problem is, that babylon 6 also renamed some flow nodes for some odd reason. We now renamed them back to the same names that the flow-parser uses, but this change will only land in 7.0

pascalduez:
This should allow us the use common imports for modules and types, which currently crashes:
```js
import { myModule, type myType } from 'my-module';
```

src/utils/__tests__/resolveToValue-test.js
```diff
@@ -49,7 +49,7 @@ describe('resolveToValue', () => {
     );
   });
 
-  it('handles SpreadProperties properly', () => {
+  it('handles SpreadPElements properly', () => {
```
fkling: Tiny typo here `SpreadElements`

fkling:
Should we just merge this? I have no problem using a beta version. We could release it as a beta/next version as well, so people could start testing it.

fkling:
I merged this into the `v3-dev` branch and will release `react-docgen@next` shortly. Even though switching to v7 shouldn't break anything, it's a big enough change that a major version bump seems appropriate.

danez:
Cool thanks

Issue #96, opened 2016-07-06
react-docgen is not compatible with react-like libraries


Chaptykov:
There are some react-like libraries such as `react-lite`, `preact-compat` and other which use API of React for components including PropTypes. All `react-docgen` methods work well with these libraries. But [this module](react-docgen disables work with react-like libraries) disallows using it. 

I have a `preact-compat` component:

```jsx
import { h, Component, PropTypes } from 'preact-compat';

export default class MyComponent extends Component {
  render() {
    return <div>{this.props.children}</div>;
  }
}

MyComponent.propTypes = {
  myProp: PropTypes.func
};
```

Actual `react-docgen` returns this value for type of myProp:

```
{
  name: 'custom',
  raw: 'PropTypes.func'
}
```

If I add `preact-compat` to whitelist `react-docgen` returns expected result:

```
{
  type: 'func'
}
```

fkling:
We could add a configuration option that lets you specify the module name(s) to validate against. Maybe we should add support for configuration files in the future if the configuration options keep increasing.

binomialstew:
@Chaptykov , I realize this is an old issue, but where is the whitelist you mentioned? I see `reactModules` in `utils/isReactModuleName`, but adding `preact`, `preact-compat`, and/or` preact/compat` to that array does not get the correct prop types to be created for me. I still see `name: 'custom',` as you described.

binomialstew:
It turns out, the `custom` name in my case was not due to `Preact` but instead because [PropTypes is not imported into each component](https://github.com/reactjs/react-docgen/issues/595).

PR #163, opened 2017-03-06
Add developit/proptypes to allowed module names

eduardoboucas:
This PR adds @developit's [proptypes module](https://github.com/developit/proptypes) to the list of recognised React modules, for better support of engines other than Facebook's React. In this case, I'm using it with [Preact](https://preactjs.com/).

Having looked at #96, I don't know if adding another hardcoded value to the list of modules is the way you want to go. If you prefer a more flexible solution of being able to supply the names of the modules to validate against, I can also look into that if you give me some pointers.

Thanks!

### DIFF
--- a/src/utils/isReactModuleName.js
+++ b/src/utils/isReactModuleName.js
@@ -10,7 +10,7 @@
  *
  */
 
-var reactModules = ['react', 'react/addons', 'react-native'];
+var reactModules = ['react', 'react/addons', 'react-native', 'proptypes'];
 
 /**
  * Takes a module name (string) and returns true if it refers to a root react


developit:
This is great. Would love to recommend react-docgen as a documentation solution that works out-of-the-box with preact :)

fkling:
Thank you for that! Overall I'm not very happy with the current solution, but I cannot think of an elegant alternative. The ideas that I had were to either add parameter that disables the "module origin" check or that allows you to define the "React" module(s). While that's flexible its also kind of annoying to have to pass such an option.  
Maybe having a configuration file for react-docgen would make this less of an issue.

If you have a suggestion I'm happy to hear about it. For now, we'll stick with what we have 😉

eduardoboucas:
Just checking if this has been merged to `npm` so I can go back to using the main package instead of my fork.

Thanks! 🤓

fkling:
Yeah, this is available in the beta version. See https://github.com/reactjs/react-docgen/releases/tag/v3.0.0-beta1

PR #164, opened 2017-03-17
-e parameter now handle regular espression

wallaroo:
if only one -e parameter is passed to cli and the value matches /.*/[gmiuy]{0,5} then this is treated as a Regular Expression

### DIFF
--- a/bin/react-docgen.js
+++ b/bin/react-docgen.js
@@ -68,6 +68,11 @@ var ignoreDir = argv.ignore;
 var excludePatterns = argv.exclude;
 var resolver;
 var errorMessage;
+var regexRegex = /\/(.*)\/([igymu]{0,5})/;
+if (excludePatterns && excludePatterns.length === 1 && regexRegex.test(excludePatterns[0])){
+  var match = excludePatterns[0].match(regexRegex);
+    excludePatterns = new RegExp(match[1], match[2]);
+}
 
 if (argv.resolver) {
   try {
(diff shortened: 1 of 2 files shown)


alexyork:
Today I discovered I need this feature, also. Can someone on the `react-docgen` team merge this into master and ship a new version?

My usecase is this: I want to point to a `/components` folder and create docs for my components but exclude all files that match the pattern `*.example.js`. Currently `-e *.example.js` sends in `['*.example.js']` as a string, not a RegEx, so I cannot achieve what I want via the CLI.

I tried this PR and passed in `-e /.example.js/` and it worked as expected.

I would love it if this got merged in :-)

fkling:
I will look at it over the weekend!

bin/react-docgen.js
```diff
@@ -68,6 +68,11 @@ var ignoreDir = argv.ignore;
 var excludePatterns = argv.exclude;
 var resolver;
 var errorMessage;
+var regexRegex = /\/(.*)\/([igymu]{0,5})/;
```
fkling: The pattern should be anchored: 

    /^\/(.*)\/([igymu]{0,5})$/;

Otherwise it looks fine to me!

fkling:
Merged as d9782c8a5d6750e35a8fc665bc446c02c9e027c7 . Will release a new version soon.

fkling:
Released v2.16.0!

PR #175, opened 2017-04-14
Revert targetting node 4 in v2.0

### LINKED ISSUE #174: Issue with transpiled code in v2.14.0 vs v2.13.0
We ran into an issue when we upgraded to v2.14.0 today that was not existing in v2.13.0.  Specifically this issue was hit when running our continuous integration night watch tests against PhantomJS.  We are seeing a console error: 'SyntaxError: Unexpected token '['' that maps to this line of code: https://github.com/reactjs/react-docgen/blob/d65bfc460d7274206c3208fc41518727acd5a490/src/utils/getMemberValuePath.js#L25  It appears that PhantomJS is not set up to use the new syntax for declaring objects.  I'm not sure what transpiling happened in v2.13.0 that is not happening in v2.14.0, but there is definitely something different between the two.

danez:
Fixes #174

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,10 +1,6 @@
 {
   "presets": [
-    ["env", {
-      "targets": {
-        "node": 4
-      }
-    }],
+    "env",
     "flow"
   ],
   "plugins": [


fkling:
Thank you!

PR #179, opened 2017-04-18
Support for create-react-class package

### LINKED ISSUE #176: Support for create-react-class for React 15.5
With [React 15.5](https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html) React.createClass was deprecated and a separate package 'create-react-class' is encouraged.  Using create-react-class currently results in
```
fixed-data-table-2/node_modules/react-docgen/dist/parse.js:60
    throw new Error(ERROR_MISSING_DEFINITION);
    ^

Error: No suitable component definition found.
```
It would be great if react-docgen supported this case.

I have [POC support here](https://github.com/wcjordan/react-docgen/commit/8c18060deeb2dd4e0d763fdc26ad9ad5b7e3a02f) and can open a PR if that seems like the right direction.  

Thanks!

wcjordan:
Fixes #176, adding support for the new create-react-class package.

I'm not familiar with recast (but have dabbled with estraverse), so any tips or suggestions are appreciated.  Thanks!

### DIFF
--- a/src/utils/isReactCreateClassCall.js
+++ b/src/utils/isReactCreateClassCall.js
@@ -21,7 +21,7 @@ var {types: {namedTypes: types}} = recast;
  * Returns true if the expression is a function call of the form
  * `React.createClass(...)`.
  */
-export default function isReactCreateClassCall(path: NodePath): boolean {
+function isReactCreateClassCallBuiltIn(path: NodePath): boolean {
   if (types.ExpressionStatement.check(path.node)) {
     path = path.get('expression');
   }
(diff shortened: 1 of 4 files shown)


fkling:
I just restarted the CI build. I failed with a call stack size error, but I don't see why it would...

wcjordan:
Weird.  I'm able to reproduce locally though, so I can dig in.

wcjordan:
I believe the issue is that this is very promiscuous
```
if (!match(path.node, {type: 'CallExpression'})) {
  return false;
}
var module = resolveToModule(path.get('callee', 'object'));
```

One heuristic I could take would be to assume the method will be named createReactClass e.g. import createReactClass from 'create-react-class'; or const createReactClass = require('create-react-class');.

Another alternative might be to make a lighter weight version of resolveToModule which doesn't follow MemberExpressions or VariableDeclarators e.g. assume we're importing directly into the variable as which we use createReactClass.

Thoughts?  Also I'd like to write some tests for these cases.  Would it be best to imitate something like isReactComponentClass-test.js?

src/utils/isReactCreateClassCall.js
```diff
@@ -32,3 +32,35 @@ export default function isReactCreateClassCall(path: NodePath): boolean {
...
+ */
+function isReactCreateClassCallModular(path: NodePath): boolean {
+  if (types.ExpressionStatement.check(path.node)) {
+    path = path.get('expression');
+  }
+
+  if (!match(path.node, {type: 'CallExpression'})) {
+    return false;
+  }
+  var module = resolveToModule(path.get('callee', 'object'));
```
fkling: This might actually be the culprit. The value of `callee` for the `CallExpression` is `Identifier`, which doesn't have an `object` field: http://astexplorer.net/#/gist/2830b1b852ac298dab2f18802133afdf/ec06e05ba93e93903bb7c85a1bd6a12622d41918 .

I think you can do `resolveToModule(path)` here, that should solve the issue.

wcjordan:
Your tip worked like a charm.  Thanks!
Also I've added tests for existing and new functionality as well as a new snapshot.
I've also verified it works for fixed-data-table-2's usage.

I'm feeling much better about things now.  Let me know if you have any additional suggestions.

fkling:
Published as v2.15.0 and v3.0.0-beta4. Thank you!

PR #194, opened 2017-06-27
Added support for render as ES7 property

### LINKED ISSUE #193: Support defining render as an ES7 property
According to [the docs](https://github.com/reactjs/react-docgen#guidelines-for-default-resolvers-and-handlers), the following must be true to detect a component:

> - When using classes, the class must either extend React.Component or define a render() method.

In some cases you might want to define your render function elsewhere, resulting in code like this:

```jsx
import view from "./view.jsx";
export default class ExampleComponent extends SomeOtherComponent {
    render = view;
}
```

It would be nice if this could be detected.

birjolaxew:
Fix #193

### DIFF
--- a/src/utils/isReactComponentClass.js
+++ b/src/utils/isReactComponentClass.js
@@ -18,10 +18,11 @@ import resolveToValue from './resolveToValue';
 var {types: {namedTypes: types}} = recast;
 
 function isRenderMethod(node) {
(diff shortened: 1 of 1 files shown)


fkling:
Merged with fe554856c9cf3250efd450a2fb4a155f69bb5600 .

Issue #187, opened 2017-06-06
Open to a resolver for styled-components?


nickbreaton:
Hello! Thanks for all the hard work on react-docgen!

I have a small issue that I would be will to help fix. The [styled-components](https://www.styled-components.com/) library does not use React syntax to generate components:

```jsx
const GreenButton = styled.p`
  color: green;
`;
```
But can still be rendered as a React component:
```jsx
<GreenButton/>
```

Would you all be open to a PR to add this as a supported type of component? Would it be better to implement this as an external library? I believe there is [demand](https://github.com/styleguidist/react-styleguidist/issues/37#issuecomment-263502454) for this with a library that depends on react-docgen.

Thanks again!

fkling:
Hey Nick!

I don't think it would beneficial to include this into the core. I wouldn't have the capacity to maintain it and it would make it harder for others (you I guess) to maintain it.

react-docgen was built with custom extensions in mind, so I think an external package would make more sense, but I'd be happy to make changes to core as necessary to support this case.

nickbreaton:
@fkling Thanks for getting back to me! I will go the route of trying to build out an extension.

Jmeyering:
@fkling 

I've gone ahead and created a custom resolver which allows for annotated export declarations to be added to a definitions list. I've used this format.

```
const MyComponent = () => (
  <h1></h1>
);

// @component
export default MyComponent;
```

Ideally this would solve this styled-components issue and allow users to annotate their exported styled-components

The issue I'm running into is that `react-docgen` validates the types of definitions that are passed to the  [`getMemberValuePath.js`][https://github.com/reactjs/react-docgen/blob/master/src/utils/getMemberValuePath.js#L35] function and limits them to a list which does not include `TaggedTemplateExpression`.

I'm willing to create a lookup method for `TaggedTemplateExpression` but was wondering if there was any information I should know before I went down that road too far.

fkling:
That's great @Jmeyering! `getMemberValuePath` is supposed find specific "members" of a component, e.g. `propTypes`.

Given a component such as 

```js
const GreenButton = styled.p`
  color: green;
`;
```

it's only possible to add additional propTypes by doing

```js
GreenButton.propTypes = ...;
```

(whether or not that is something that is usually done is irrelevant for this discussion)

Maybe it would make sense to default any unknown expression to `getMemberExpressionValuePath` (which would have to be updated to support arbitrary expressions as well) ?

Jmeyering:
[...]

PR #195, opened 2017-07-13
Allow getMemberValuePath to get data from a TaggedTemplateExpression

Jmeyering:
This will enable additional tooling like styled-components and diet-cola to create components which get 'picked up' by react-docgen.

Addresses #187

### DIFF
--- a/src/utils/getMemberExpressionValuePath.js
+++ b/src/utils/getMemberExpressionValuePath.js
@@ -36,7 +36,8 @@ function resolveName(path) {
 
   if (
     types.FunctionExpression.check(path.node) ||
-    types.ArrowFunctionExpression.check(path.node)
+    types.ArrowFunctionExpression.check(path.node) ||
+    types.TaggedTemplateExpression.check(path.node)
   ) {
     if (!types.VariableDeclarator.check(path.parent.node)) {
       return; // eslint-disable-line consistent-return
--- a/src/utils/getMemberValuePath.js
+++ b/src/utils/getMemberValuePath.js
@@ -36,6 +36,7 @@ function isSupportedDefinitionType({node}) {
   return types.ObjectExpression.check(node) ||
     types.ClassDeclaration.check(node) ||
     types.ClassExpression.check(node) ||
+    types.TaggedTemplateExpression.check(node) ||
 
     // potential stateless function component
     types.VariableDeclaration.check(node) ||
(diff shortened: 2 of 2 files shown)


fkling:
Thank for this! I think this is fine overall, but please add a test.

It might also be worth to add a comment why this was added and maybe add a link to your resolver.

Once this all works together, we should list your resolver somewhere, maybe in the wiki.

Jmeyering:
Any direction on how to add a test? The test suite by default uses the built in resolvers, which do not include tagged template literals.

I don't see any place where I can unit test the individual methods that I changed. I will gladly add a comment to my changes.

Jmeyering:
@fkling 

A couple things

1. I checked `react-styleguidist` and it looks as though they are locked to v3-dev for their dep on `react-docgen` As that is the primary use case that I am aware of at this time, I have made a new PR for the `v3-dev` branch #198. Should we close this PR and only work on this as a v3-dev feature?
2. I've added inline documentation for my changes in #198.
3. Testing is still confusing me a bit. Any direction would be greatly appreciated

Jmeyering:
@fkling Is there any one who could provide some direction on getting testing implemented? As I've commented before it doesn't seem to me that there are any unit tests explicitly covering the methods that I have altered.  Fixtures are used to test most of the functionality of react-docgen. But as the library itself does not have any way to resolve a TaggedTemplateLiteral as a component, I'm a bit stuck. 

Thanks so much for your time.

fkling:
Sorry for letting this sit for so long (I also was away in between)....

Regarding testing, you could start by adding tests to `getMemberExpressionValuePath-test.js` and `getMemberValuePath-test.js`. Something like

```js
 describe('TaggedTemplateLiteral', () => {
    it('finds "normal" property definitions', () => {
      var def = statement(`
        var Foo = foo\`bar\`
        Foo.propTypes = {};
      `);

      expect(getMemberExpressionValuePath(def, 'propTypes'))
        .toBe(def.parent.get('body', 1, 'expression', 'right'));
    });
});
```

and

```js
  it('handles TaggedTemplateLiterals', () => {
    var path = expression('foo``');

    getMemberValuePath(path, 'foo');
    expect(getPropertyValuePath).toBeCalledWith(path, 'foo');
  });
```

Jmeyering:
Closing this as it's merged into v3-dev

PR #211, opened 2017-08-24
feat(resolve): resolve Object.keys() in PropType.oneOf()

### LINKED ISSUE #153: support `Object.keys` within `oneOf`
I'd like to use `Object.keys` to generate the list of values possible for `oneOf` but have found that this currently does not work. Has anyone run into this before? Not sure if this is a feature request or if there is a better way I could be writing my propType.

danez:
Creates a new util to extract keys from `Object.keys()` and `ObjectExpressions` when used inside of `oneOf()`
It also follows spreads and tries to resolve them. Currently it bails out if either a `Property` is computed and the computed key is not a `Literal` or if a spread cannot be resolved.

Getters will be handled the same as normal properties, but setters will be ignored. Not sure if this makes 100% sense, but this can be easily changed.

This fixes #153.

As a follow up the same can be done easily for `Object.values`.

!! When this gets merged to v3, `SpreadProperty` needs to be replaced with `SpreadElement` !!

### DIFF
--- a/src/utils/__tests__/getPropType-test.js
+++ b/src/utils/__tests__/getPropType-test.js
@@ -236,8 +236,10 @@ describe('getPropType', () => {
 
       expect(getPropType(propTypeExpression)).toEqual({
         name: 'enum',
-        value: 'Object.keys(TYPES)',
-        computed: true,
+        value: [
+          {value: '"FOO"', computed: false},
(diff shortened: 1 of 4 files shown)


stepancar:
@danez

stonebk:
It would be great if this could be done for `Object.values()` as well as @danez mentioned.

Issue #213, opened 2017-08-30
ERROR_MULTIPLE_DEFINITIONS


konsumer:
I'd like for it to be ok to have multiple exports in my interface. I tend to do this sort of thing:

```js
import React from 'react'
import { connect } from 'react-redux'
import PropTypes from 'prop-types'

/** A dumb component that says "hello" */
export const Docs = ({name}) => (<div>Hello {name}</div>)

Docs.propTypes = {
  /** the name of the person to say "hello" to */
  name: PropTypes.string
}

Docs.defaultProps = {
  name: 'World'
}

export const mapStateToProps = state => ({name: state.name})

export default connect(mapStateToProps)(Docs)
```

This makes it easier to unit-test my redux state-mapper & standalone Component (without redux) and allows me to make storybook stories that don't need redux (I just hand them a snapshot of the relevant state-pieces in the story) while still being really easy to work with in my redux app.

If I run react-docgen on the above file, I get this error:
```
Error with path "Docs.js": Error: Multiple exported component definitions found.
Error: Multiple exported component definitions found.
```

If I comment out those and loop over the definitions, I get great output:

```json
{"description":"A dumb component that says \"hello\"","methods":[],"props":{"name":{"type":{"name":"string"},"required":false,"description":"the name of the person to say \"hello\" to","defaultValue":{"value":"'World'","computed":false}}}}
```

Does this part serve a purpose that makes this solution not a good one? I have a feeling I am missing some key part of the original functionality here.

fkling:
Since you know about `src/resolver/findAllExportedComponentDefinitions.js` now, I guess we can close this?

konsumer:
Yep, looks like I need to work on the babel-loader, and set it up to use `findAllExportedComponentDefinitions`. Sorry for the noise.

fkling:
No worries!

PR #214, opened 2017-08-30
no error on multiple definitions (#213)

konsumer:
Related to #213 : multiple exported definitions are ok.


danez:
Thanks for the PR.
You could probably solve your problem by using one of the other resolvers react-docgen ships with:
https://github.com/reactjs/react-docgen/tree/master/src/resolver
Also described in the CLI docs: https://github.com/reactjs/react-docgen#cli

konsumer:
Ah, yes, I like this a lot better. Maybe I should try to get storybook to use a different-than-default resolver. Feel free to close this, sorry for the noise.

PR #218, opened 2017-09-24
Add support for function paramets without names

odinuge:
Add support for flow functions with parameters that doesn't have a name.
Eg. `value: string => void` and `value: (string, number) => void`

This kind of functions ins't currently supported by `react-docgen`, and it's unable to parse the file.

Source: [flow](https://flow.org/en/docs/types/functions/#toc-function-types)

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,7 @@ type Props = {
   literalsAndUnion: 'string' | 'otherstring' | number,
   arr: Array<any>,
   func?: (value: string) => void,
+  noParameterName?: string => void,
   obj?: { subvalue: ?boolean },
 };
 
@@ -282,6 +283,20 @@ we are getting this output:
         "signature":{
           "arguments":[
             { "name":"value", "type":{ "name":"string" } }
+          ],
+          "return":{ "name":"void" }
+        }
+      },
+      "required":false
+    },
+    "noParameterName":{
+      "flowType":{
+        "name":"signature",
+        "type":"function",
+        "raw":"string => void",
+        "signature":{
+          "arguments":[
+            { "name":"", "type":{ "name":"string" } }
           ],
           "return":{ "name":"void" }
         }
--- a/src/utils/__tests__/getFlowType-test.js
+++ b/src/utils/__tests__/getFlowType-test.js
@@ -139,6 +139,27 @@ describe('getFlowType', () => {
     }, raw: '(p1: number, p2: ?string) => boolean'});
   });
 
+
+  it('detects function signature types without parameter names', () => {
+    var typePath = expression('x: (number, ?string) => boolean').get('typeAnnotation').get('typeAnnotation');
+    expect(getFlowType(typePath)).toEqual({name: 'signature', type: 'function', signature: {
+      arguments: [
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -288,6 +289,20 @@ we are getting this output:
...
       "required":false
     },
+    "noParameterName":{
+      "flowType":{
+        "name":"signature",
+        "type":"function",
+        "raw":"string => void",
+        "signature":{
+          "arguments":[
+            { "name":"_", "type":{ "name":"string" } }
```
danez: What do you think about instead of using `'_'`, setting the field to `null`? It seems clearer to me that a consumer of react-docgen checks if the `name === null` instead of a "random" string `name === '_'`

odinuge: Hi @danez,

Got the inspiration from a flow-addon for atom (see screenshot). But I agree that `null` would be a better and cleaner value for working with the data afterwards. Will update the PR.

screenshot from 2017-10-05 22-16-02

odinuge: Edit: `getNameOrValue` returns a string value, and therefore `null` would break flow. What do you think @danez?

danez: Hmm yes seems it is used quite a bit. We could maybe return an empty string? Then `!name` would work in the resulting documentation.
My intention is just that we should return something else than a non-empty string, because in the `'_'` case you would not be able to differ if the variable name was `_` or empty.

odinuge: Updated the PR setting the name to an empty string now.

src/utils/getNameOrValue.js
```diff
@@ -25,6 +25,8 @@ export default function getNameOrValue(path: NodePath, raw?: boolean): string {
       return node.name;
     case types.Literal.name:
       return raw ? node.raw : node.value;
+    case types.FunctionTypeParam.name:
```
fkling: Doing this here seems weird to me. I wouldn't expect `getNameOrValue` to be called with with a type annotation (but who knows...). It's also not quite right, since some `FunctionTypeParam` do have a name.

Personally I feel it makes more sense to fix this in `getFlowType`. If the name of the `FunctionTypeParam` is `null`, assign an empty string instead. Something like

    name: param.node.name ? param.node.name.name || '';

@danez , why does [this](https://github.com/reactjs/react-docgen/blob/67fcd097705eff40dac17c384353d64c3add8daa/src/utils/getFlowType.js#L151) use `getProperty` to extract the name at the moment? Shouldn't it directly by `getNameOrValue(param.get('name'))` or just `param.node.name.name`? 
Am I looking at the wrong syntax tree or am I missing a case here?

danez: Good question, I have no idea :) But seems unnecessary as it can never be computed or be a literal. 

Doing `name: param.node.name ? param.node.name.name || '';` should be correct in `getFlowType` unless the tests think different.

odinuge:
Hi @fkling,
After looking further into `getFlowType`, it makes more sense to do it your way with a simple inline ternary operator, instead of using the `getNameOrValue`. Updated the PR, and the tests still works as expected. :smile:

Issue #221, opened 2017-10-16
Make props with infered default types from `.defaultProps` optional?


magicmark:
Hi! Apologies if this has already been raised, but when we add default types to a component as such:

```
// @flow

import React from 'react';

type Props = {
    a: string,
    b: string,
};

const Foo = (props: Props) => <div>Hello World! {props.a} {props.b}</div>;

Foo.defaultProps = {
    a: 'foo',
    b: 'bar',
};

export default Foo;
```

This implies that any props that are in `defaultProps` are _optional_.

However react-docgen still sees them as `"required": true`:

```
$ node_modules/.bin/react-docgen src/components/Foo/Foo.js | jq .
{
  "description": "",
  "methods": [],
  "props": {
    "a": {
      "flowType": {
        "name": "string"
      },
      "required": true,
      "description": "",
      "defaultValue": {
        "value": "'foo'",
        "computed": false
      }
    },
    "b": {
      "flowType": {
        "name": "string"
      },
      "required": true,
      "description": "",
      "defaultValue": {
        "value": "'bar'",
        "computed": false
      }
    }
  }
}
```

(using `react-docgen@3.0.0-beta8`)

Thanks!

fkling:
How does flow behave when type checking this component? Does it take default props into account?

magicmark:
@fkling flow (0.56.0) shows the following 

```
$ node_modules/.bin/flow gen-flow-files Foo.js
// @flow

import React from "react";
declare export default function(props: {a: string, b: string}): $Exact<{+key: string | number | null, +props: mixed /* UNEXPECTED TYPE: EvalT */, +ref: any, +type: ElementType}>;
```

This problem shows up when we show our component on styleguidist (which uses react-docgen): https://i.imgur.com/6u1GrRR.png

fkling:
OK.... the output doesn't tell me much but I tried it on https://flow.org/try and flow doesn't complain about missing props when they have default values, so yeah, we should mark them as optional.

adityavohra7:
[...]

PR #227, opened 2017-10-26
Added post-processing step for documentation objects

adityavohra7:
Definitely a WIP, but thought I'd get a PR going to have something to discuss on. This branch adds a post-processing step for documentation objects. Currently, this post-processing step only does what's needed to "fix" #221, but I can imagine we could use it for validation of documentation objects as well? I haven't added any tests yet. Let me know what you think!

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,8 +19,8 @@
   "scripts": {
     "build": "rimraf dist/ && babel src/ --out-dir dist/ --ignore __tests__,__mocks__",
     "lint": "eslint src/ bin/",
-    "prepublish": "yarn run build",
-    "preversion": "yarn run lint",
+    "prepublish": "yarn build",
+    "preversion": "yarn lint",
     "test": "jest",
     "test:ci": "yarn lint && yarn flow && yarn test --runInBand",
     "watch": "yarn build --watch"
--- a/src/__tests__/__snapshots__/main-test.js.snap
+++ b/src/__tests__/__snapshots__/main-test.js.snap
@@ -38,6 +38,7 @@ Object {
         "computed": false,
         "value": "\\"primary\\"",
       },
+      "required": false,
     },
   },
 }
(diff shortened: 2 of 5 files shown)


adityavohra7:
Looking into the test failures!

adityavohra7:
Friendly bump! Let me know if you think we should take some other approach here!

rosskevin:
@adityavohra7 is this still a WIP or is it good to use/merge?

@danez #221 is a huge problem for material-ui docs - if you could take a look and let us know if this direction is good, we need to get our docs generating correctly again.  

Thanks to you both!

adityavohra7:
@rosskevin, the PR definitely needs a couple of tests. I was waiting for any maintainer to look it over though to confirm they agreed with the approach the PR takes. Majority of the doc generation is "independent" (handlers are unaware of each other), but this approach sort of "combines" the documentation that each handler creates.

rosskevin:
I just noticed that this is for `master` and we are already dependent on the `3.0-dev` releases.  I'm going to try and work this code into that branch and see if I can get material-ui running again.

adityavohra7:
Friendly bump! Any input by other contributors/maintainers would be appreciated!

danez:
Sorry for the late reply, this looks overall good, to me. The only thing I#m not sure is that this right now would hardcode the behaviour. I don#t see a reason why anyone does not want to have this, but who knows. So maybe it would be better to make it a prober handler and add it as the last one. I guess that should work too?
Some tests would be good.

And targetting the master branch is totally fine, I can pick stuff to the v3 branch after it is merged as long as there are tests that verify it works. :)

Nice work.

adityavohra7:
Hi @danez! Thanks for the review! I've rebased on master and added a simple test case for a component with defaultProps and flow type annotations for props.

The only reason I didn't add the "combining of defaultProps and flow types" behavior in the form of a handler is that there'd be an implicit requirement for this handler to run last (because it depends on the results of the props and flow handlers). If we prefer that, I don't mind changing the current post-processing step to a handler!

adityavohra7:
Friendly bump! @danez, what do you think about the updated PR? 😊

danez:
I merge this now as is, I think having it hardcoded is okay for now. We will see if someone complains about wanting to disable this behaviour. If so we can still convert it to a handler without breaking applications.
Thanks for your work.

PR #231, opened 2017-11-18
Post process flow default props

### LINKED ISSUE #221: Make props with infered default types from `.defaultProps` optional?

rosskevin:
Fixes #221 

Based on the work by @adityavohra7 in #227 (on `master`), I have applied a similar change to `v3-dev` with post processing of defaultProps.  Props are not required when they are specified in `defaultProps`.   

Further, I added `component_11.js` as a diverse test case taken from `material-ui`.  It Is an all flow test that exercises required, optional, and parameterized types.

Use case: https://github.com/callemall/material-ui/issues/9222

### DIFF
--- a/src/__tests__/fixtures/component_11.js
+++ b/src/__tests__/fixtures/component_11.js
@@ -35,6 +35,11 @@ export type Props = {
    */
   elevation: number,
   /**
+   * Useful to customize the rows per page label. Invoked with a `{ from, to, count, page }`
+   * object.
+   */
+  labelRowsPerPage: Node,
+  /**
    * Transition component.
    */
   transition: ComponentType<*>,
@@ -42,6 +47,7 @@ export type Props = {
 
 class Paper extends React.Component<ProvidedProps & Props> {
   static defaultProps = {
+    labelRowsPerPage: ('Rows per page:': Node),
     component: ('div': ElementType),
     elevation: 2,
   };
--- a/src/__tests__/fixtures/component_11.js
+++ b/src/__tests__/fixtures/component_11.js
@@ -16,6 +16,12 @@ export type Origin = {
   vertical: 'top' | 'center' | 'bottom' | number,
 };
 
+export const duration = {
+  standard: 300,
+};
+
+export type TransitionDuration = number | { enter?: number, exit?: number } | 'auto';
+
 export type Props = {
   /**
    * Other base element props.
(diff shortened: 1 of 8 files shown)


rosskevin:
Hold, I have a bug to work out.

rosskevin:
Fix pushed.  Good to go.

rosskevin:
Ok, this is good to go and passes with the material-ui codebase.

src/__tests__/fixtures/component_11.js
```diff
@@ -0,0 +1,101 @@
...
+};
+
+class Paper extends React.Component<ProvidedProps & Props> {
+  static defaultProps = {
+    anchorOrigin: ({
+      vertical: 'top',
+      horizontal: 'left',
+    }: Origin),
+    labelRowsPerPage: ('Rows per page:': Node),
+    component: ('div': ElementType),
```
danez: Out of curiosity: What is this doing?

danez:
I posted already in #227, seems this fixes some printing for casts too though

src/__tests__/fixtures/component_11.js
rosskevin: Casting to ElementType to avoid other flow errors and avoid typing all default props. Just one alternative.  BTW We have since removed flow from our codebase.

danez: I see, thanks for explaining.

Yeah my company also dropped flow, because it was to hard to keep up and we had to add lots of workarounds.

danez:
#227 has been merged and I picked the other changes here: a893dec3ea9ef565a029357d70e7cc496d0cb77c

PR #233, opened 2017-11-21
add support for prop-types package

dickson-tong:


dickson-tong:
sorry mistake. wrong repo

PR #237, opened 2017-12-16
[RFC] Speed up react-docgen by 25-40%

sophiebits:
I did some profiling and noticed that recast has a significant amount of overhead due to the fact that it copies every tree it parses and has complicated printing logic. I wasn't sure if react-docgen uses any mutative recast features so I tried to remove it.

The good news: this is about 25-40% faster in my testing (40% faster when run on 20 files, 25% faster when run on 700 components from FB internal repos -- probably the JIT mitigates some of the slowness of recast).

The bad news: this doesn't actaully work, because (surprise, surprise) we actually do use some of the modification features of recast. When these are used, we don't know how to reprint.

```
$ ag --ignore __tests__ builders\\.
src/utils/normalizeClassDefinition.js
79:            var classProperty = builders.classProperty(

src/utils/resolveObjectKeysToArray.js
98:        .map(value => builders.literal(value))
100:      return new NodePath(builders.arrayExpression(nodes));

src/utils/resolveToValue.js
33:        builders.memberExpression(
```

Maybe we could reimplement a small subset of the printing logic? Or maybe that's too impractical. FWIW all but one of the 700 files I tested produced the exact same output including whitespace in the `raw` JS source fields.

### DIFF
--- a/src/parse.js
+++ b/src/parse.js
@@ -50,7 +50,11 @@ export default function parse(
   resolver: Resolver,
   handlers: Array<Handler>
 ): Array<Object>|Object {
-  var ast = recast.parse(src, {esprima: babylon});
+  let ast = babylon.parse(src, {
+   sourceType: 'module'
+  });
+  ast._docgenOriginalFileSource = src;
+  ast = recast.types.builders.file(ast, null);
   var componentDefinitions = resolver(ast.program, recast);
 
   if (Array.isArray(componentDefinitions)) {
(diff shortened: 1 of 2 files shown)


sophiebits:
(FB-only context for why I ended up looking at this: https://fburl.com/zbfp3cl0.)

danez:
Closing this now as I created #349 based on your idea.

The reprinting was actually only failing in the case of `resolveObjectKeysToArray` and `resolveObjectKeysToArray` where we create literals with builders, all the other cases are either never reprinted or contain nodes from the original AST.

Thank you very much.

Out of curiosity what was the FB internal reason to look into this? Is it too slow?

PR #240, opened 2018-01-04
Update recast for flow type variance

TheSavior:
Update recast so that it has the proper ast-types for covariant flow types.

PR #241, opened 2018-01-05
Support flow type spread

TheSavior:
Supporting type spread props:

```
type Props = {|
  /**
   * Spread props potentially from another component
   */
  ...OtherProps,

  /**
   * The first prop
   */
  prop1: string,
|}
```

This is a stacked PR on top of https://github.com/reactjs/react-docgen/pull/240


TheSavior:
It seems like we should change the `getFlowType` function to return something special for ObjectTypeSpread.

danez:
I merged #240, could you please rebase?

TheSavior:
Rebased. This PR was a little less obvious to me what the right approach was. I tried to follow the algorithm conventions that existed but I'm not sure if the output of the spread type is what it actually should be. I would appreciate a bit closer review of those pieces.

src/utils/__tests__/getFlowType-test.js
```diff
@@ -232,6 +232,23 @@ describe('getFlowType', () => {
...
+    var typePath = statement(`
+      var x: { ...OtherProps, a: string };
+    `).get('declarations', 0).get('id').get('typeAnnotation').get('typeAnnotation');
+
+    expect(getFlowType(typePath)).toEqual({
+      name: 'signature',
+      type: 'object',
+      signature: {
+        properties: [
+          { key: 'OtherProps', value: { name: 'OtherProps', required: true } },
```
danez: I think it is not really clear what this is just from the json alone. It could be also interpreted as 
```js
type a = { OtherProps: OtherProps };
```
What i would try to do is resolve `OtherProps` and if it resolves to an other ObjectType within the same file then I would just inline this in the result. We do the same when resolving stuff here https://github.com/reactjs/react-docgen/blob/master/src/utils/resolveObjectKeysToArray.js#L59..L67

In the case when it cannot be resolved I'm not sure what the best approach is. We could either ignore the spread completely then, which we currently do for regular non-type ObjectSpread. Or somehow add a special field in the output aka "inherits" (or some better name) which just contains the names of the Type that was spreaded into the object. If I think about it spreading types is just some kind of multi inheritance and it is probably interesting to know. 

```js
{
  name: 'signature',
  type: 'object',
  includes/inherits/implements/??: [ 'OtherProps' ],
  signature: {
    properties: [
      { key: 'a', value: { name: 'string', required: true } },
    ]
  }
}
```

Not sure if that makes sense though.

TheSavior: The logic for resolving the local types is already being done. You can see it in action in the component snapshot test, prop `OtherLocalProps`.

Yeah, the real question is what to do when the type is imported from another file, 90% of our use case internally. We definitely don't want to ignore the spread. I think for us at least, saying what the types are is pretty explanatory. It's typically things like `...AccessibilityProps`. 

An explicit inherits field is interesting. I am not familiar enough with this project and how we consume the output to know what is ideal here. I think I defer to you.

src/__tests__/__snapshots__/main-test.js.snap
```diff
@@ -267,6 +267,35 @@ Object {
...
   "methods": Array [],
   "props": Object {
+    "OtherComponentProps": Object {
+      "description": "Spread props from another file",
+      "flowType": Object {
+        "name": "OtherComponentProps",
+      },
+      "required": false,
+    },
+    "OtherLocalProps": Object {
```
TheSavior: It seems like this could be confused with a prop called `OtherLocalProps` that has these specific keys. However, since it is being spread in, the properties of OtherLocalProps should probably be flattened into the top level props? Is that correct?

src/utils/__tests__/getFlowType-test.js
TheSavior: This is an example of a component where we type spread props in and how we'd probably expect them to show up in docs. https://facebook.github.io/react-native/docs/touchableopacity.html#props

TheSavior:
@danez, do you have any more thoughts on these open questions?

TheSavior:
I'm thinking that we want to distinguish between the spread properties and the normal ones. Also, since we want to resolve local types and possibly at some point have a way to reference where the definition originally comes from, we want the inherited list to be objects instead of plain strings.

I'm thinking this structure:

```
  it('handles ObjectTypeSpreadProperty', () => {
    var typePath = statement(`
      var x: { ...OtherProps, ...MyType, a: string };

      type MyType = { myString: string };
    `).get('declarations', 0).get('id').get('typeAnnotation').get('typeAnnotation');

    expect(getFlowType(typePath)).toEqual({
      name: 'signature',
      type: 'object',
      signature: {
        inheritedProperties: [
          {name: 'OtherProps'},
          {name: 'MyType', value: {
            key: 'myString', value: { name: 'string', required: true } },
          },
        ],
        properties: [
          { key: 'OtherProps', value: { name: 'OtherProps', required: true } },
          { key: 'a', value: { name: 'string', required: true } },
        ],
      },
      raw: '{ ...OtherProps, ...MyType, a: string }'});
  });
```

src/__tests__/__snapshots__/main-test.js.snap
```diff
@@ -267,6 +267,35 @@ Object {
   "displayName": "MyComponent",
   "methods": Array [],
   "props": Object {
+    "OtherComponentProps": Object {
```
TheSavior: Even though my getFlowType test is working, these are still being put under props in the snapshot. Trying to track this down.

src/__tests__/__snapshots__/main-test.js.snap
```diff
@@ -267,6 +267,35 @@ Object {
...
+      "description": "Spread props from another file",
+      "flowType": Object {
+        "name": "OtherComponentProps",
+      },
+      "inherited": true,
```
TheSavior: These have a flag for whether it is inherited. This was cleaner than adding a new top level key to the output documentation.

TheSavior:
Alright, I think this is good to go now.

TheSavior:
@danez, kind bump. @fkling is on vacation so I can't bug him internally to get this merged and released. :)

danez:
Why did you decide to do it in 2 different ways? Once with `inheritedProps` and once with a `inherited` flag?

TheSavior:
I could have gone either way, I actually initially had them both as a separate `inherited` object. The reason I didn't was partly because the code necessary to add inherited props to a new `inheritedProps` field in the documentation class is pretty hairy. Adding an inherited flag there was the simplest thing to do and required the most surface level changes that would be most likely to be merged. I figured if we wanted to split them out later it would be easier going this direction than the other direction.

There are a few different reasons I made `getFlowType` have a separate `inheritedProperties`. Fundamentally, the output from this function seems different than the component documentation output. One is describing the props, one is describing the type of a generic variable in JavaScript. The `properties` array also had `key` whereas the key of an inherited property doesn't actually make sense so `inheritedProperties` use `name`. 

All in all, I don't feel strongly either direction. I'm happy to change approach if someone with more context has a preference. I mostly just want to get Facebook's UIDocs to stop blowing up on these new flow types. Perhaps we can merge and ship a PR that does that and then we can figure out the best output for these additional types.

src/__tests__/__snapshots__/main-test.js.snap
TheSavior: Move this to composes

src/__tests__/__snapshots__/main-test.js.snap
```diff
@@ -267,6 +267,35 @@ Object {
...
+    },
+    "OtherLocalProps": Object {
```
TheSavior: Inline this into props.

src/handlers/flowTypeHandler.js
```diff
@@ -9,22 +9,40 @@
...
+    type = getFlowType(path.get('value'));
+    propDescriptor.required = !path.node.optional;
+  }
+
   if (type) {
     propDescriptor.flowType = type;
-    propDescriptor.required = !path.node.optional;
   }
+
+  setPropDescription(documentation, path);
```
TheSavior: Add a comment to this.

src/__tests__/__snapshots__/main-test.js.snap
```diff
@@ -263,10 +263,35 @@ Object {
...
 exports[`main fixtures processes component "component_11.js" without errors 1`] = `
 Object {
+  "composes": Array [
+    "OtherProps",
+  ],
   "description": "",
   "displayName": "MyComponent",
   "methods": Array [],
   "props": Object {
+    "OtherLocalProps": Object {
```
TheSavior: This is wrong. Not spread in far enough.

src/__tests__/fixtures/component_11.js
```diff
@@ -14,7 +14,23 @@
 
 import React from 'react';
 
+import { OtherComponentProps as OtherProps } from 'NonExistentFile';
+
+type OtherLocalProps = {|
+  fooProp?: string,
```
TheSavior: Add and pull in a docblock comment here.

danez:
Looks good to me now. So you now changed it to use `composes`? That sounds really good to me. Should we try to align the other PR? `composes` and `importedProperties`... or do you think this are different usecases?

TheSavior:
I think the other one needs a bit more work still. All of the inline comments I left were from doing an in person code review with @fkling. This PR should be good to go though. I’ll try to get Felix to review it closer today. :-)

src/handlers/flowTypeHandler.js
```diff
@@ -17,13 +18,33 @@ import getPropertyName from '../utils/getPropertyName';
...
+    const name = path.get('argument').get('id').get('name');
+    const resolvedPath = resolveToValue(name);
+
+    if (resolvedPath && types.TypeAlias.check(resolvedPath.node)) {
+      const right = resolvedPath.get('right');
+      applyToFlowTypeProperties(right, propertyPath => {
+        setPropDescriptor(documentation, propertyPath)
+      });
+    } else {
+      documentation.addComposes(name.node.name);
```
fkling: So this takes the variable name used in the spread syntax, right? Did we plan to use the module name instead at some point? I can't remember.

TheSavior: I think where it ended up was that it was too difficult to make something meaningful here.

```
import type { MyType } from 'MyFile';

type MyFileMyProperty = $PropertyType<MyType, 'myProperty'>;

type Props = {|
  ...MyFileMyProperty
|}
```

It doesn't make sense to say that it composes `MyFile.MyFileMyProperty`, in this case it would be `MyFile.MyType.myProperty` but the logic can get exceedingly complicated and it doesn't make sense for us to try to understand it statically. I think where we landed was to just use the alias, if they want it to be more explicit in the generated docs they can use a more explicit alias. 

This still leaves us open to improving it at a future point though.

fkling: Alright

PR #242, opened 2018-01-09
Support deep objects with spread flow types

TheSavior:
This handles props where a value in that type is a object which contains spread types.

Requires the changes in #241

### DIFF
--- a/src/utils/__tests__/getFlowType-test.js
+++ b/src/utils/__tests__/getFlowType-test.js
@@ -232,6 +232,39 @@ describe('getFlowType', () => {
     }, raw: '{ a: string, b: ?xyz }'});
   });
 
+  it('handles ObjectTypeSpreadProperty', () => {
+    var typePath = statement(`
+      var x: { ...OtherProps, ...MyType, a: string };
+
+      type MyType = { myString: string };
+    `).get('declarations', 0).get('id').get('typeAnnotation').get('typeAnnotation');
+
+    expect(getFlowType(typePath)).toEqual({
+      name: 'signature',
+      type: 'object',
+      signature: {
+        inheritedProperties: [
+          { name: 'OtherProps', value: { name: 'OtherProps' } },
+          { name: 'MyType',
+            value: {
+              name: 'signature',
+              type: 'object',
+              signature: {
+                properties: [
+                  { key: 'myString', value: { name: 'string', required: true } },
+                ],
+              },
+              raw: '{ myString: string }',
+            },
+          },
+        ],
+        properties: [
+          { key: 'a', value: { name: 'string', required: true } },
+        ],
+      },
+      raw: '{ ...OtherProps, ...MyType, a: string }'});
+  });
+
   describe('React types', () => {
     function test(type, expected) {
       var typePath = statement(`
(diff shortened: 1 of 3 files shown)


src/utils/__tests__/getFlowType-test.js
```diff
@@ -232,6 +232,39 @@ describe('getFlowType', () => {
...
+      type MyType = { myString: string };
+    `).get('declarations', 0).get('id').get('typeAnnotation').get('typeAnnotation');
+
+    expect(getFlowType(typePath)).toEqual({
+      name: 'signature',
+      type: 'object',
+      signature: {
+        inheritedProperties: [
+          { name: 'OtherProps', value: { name: 'OtherProps' } },
+          { name: 'MyType',
```
TheSavior: This should be inlined.

src/utils/__tests__/getFlowType-test.js
```diff
@@ -232,6 +232,39 @@ describe('getFlowType', () => {
...
+
+      type MyType = { myString: string };
+    `).get('declarations', 0).get('id').get('typeAnnotation').get('typeAnnotation');
+
+    expect(getFlowType(typePath)).toEqual({
+      name: 'signature',
+      type: 'object',
+      signature: {
+        inheritedProperties: [
+          { name: 'OtherProps', value: { name: 'OtherProps' } },
```
TheSavior:
```
import type { BarType as Foo } from 'bar';
```

```
{ ...Foo }
```

```
{ 
module: 'bar',
name: 'BarType'
}
```


---------------

```
import type URI from 'URI';
```

```
{ ...URI }
```

```
{ 
module: 'URI',
name: 'default'
}
```

src/utils/__tests__/getFlowType-test.js
```diff
@@ -232,6 +232,39 @@ describe('getFlowType', () => {
...
+      var x: { ...OtherProps, ...MyType, a: string };
+
+      type MyType = { myString: string };
+    `).get('declarations', 0).get('id').get('typeAnnotation').get('typeAnnotation');
+
+    expect(getFlowType(typePath)).toEqual({
+      name: 'signature',
+      type: 'object',
+      signature: {
+        inheritedProperties: [
```
TheSavior: importedProperties

src/utils/__tests__/getFlowType-test.js
danez: I like that. Not sure if the local identifier is worth having in here too, 

```js
{ 
module: 'bar',
name: 'BarType',
local: 'Foo'
}
```

src/utils/__tests__/getFlowType-test.js
danez: That makes sense, because the ultimate goal of react-docgen is to describe the props of react components and not the complete AST of js files. So as long as we have all the information we should add them to properties I think.

src/utils/__tests__/getFlowType-test.js
danez: or compose like in the flow PR? From a purely react perspective, I think `composes` makes a lot of sense. When describing the props it could (very unlikely) also be a global that gets spread into the props, where this name then does not make sense. (Btw we should add a test for spreading a global variable, which has no declaration in the scope)

src/utils/__tests__/getFlowType-test.js
fkling: @danez: I originally thought that anything that's local wouldn't be of interest to someone reading the documentation. But who knows, maybe it's useful for the person that generates the documentation.

PR #243, opened 2018-01-11
Ignore ObjectTypeSpreadProperty

TheSavior:
This is the simplest way to ignore object type spread properties and have react-docgen not blow up. 

Without this change an exception would throw saying `Propery name must be an Identifier or a Literal`. 

Since this PR ignores the spread types, they don't show up in the output. There are other PRs to add that functionality once we figure out the ideal output structure: #241 #242

### DIFF
--- a/src/__tests__/fixtures/component_11.js
+++ b/src/__tests__/fixtures/component_11.js
@@ -14,7 +14,23 @@
 
 import React from 'react';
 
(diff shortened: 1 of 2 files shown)


TheSavior:
cc @danez

fkling:
Not needed anymore, right?

PR #254, opened 2018-02-21
Update eslint and related deps

adityavohra7:
- updated eslint and babel-eslint
- changed .eslintrc to .eslintrc.js because it's [preferred](https://eslint.org/docs/user-guide/configuring#configuration-file-formats)


danez:
Thank you

PR #261, opened 2018-03-20
Add ability to detect "required" on shapes

NikhilVerma:
This is something that's causing us troubles when creating our styleguide. React Docgen does not detect shapes or enums as required or not.

This PR adds the support. What I've done

- Added and updated tests
- Made sure lint rules were fine

It's my first time contributing to this repo so please let me know If I am doing something wrong.

Thanks! 👋

### DIFF
--- a/src/__tests__/__snapshots__/main-test.js.snap
+++ b/src/__tests__/__snapshots__/main-test.js.snap
@@ -73,6 +73,7 @@ Object {
       "type": Object {
         "computed": true,
         "name": "shape",
+        "required": true,
         "value": "Child.propTypes",
       },
     },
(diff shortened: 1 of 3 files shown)


src/__tests__/__snapshots__/main-test.js.snap
```diff
@@ -81,6 +82,7 @@ Object {
       "required": true,
       "type": Object {
         "name": "shape",
+        "required": true,
```
danez: Can you elaborate in which example exactly it is not adding the required field? Because in this example it is now twice, because we already add the required field in the prop object. see line 82.

NikhilVerma:
@danez  Good point, I'll have a look why it needs to be duplicated.

NikhilVerma:
```
if (propInfo.defaultValue) {
      propInfo.required = false;
    }
```

@danez  Sorry I'll close this PR, seems like this is by design. If a propType has a default value it can't be forced as required. Which I guess makes sense, I'll update our doc generator to keep this in mind.

PR #268, opened 2018-05-24
Allow for programmatic usage of "react-docgen" binary features

indexzero:
Howdy! Doing some fancy(ish) traversals of a bunch of `react` components using `react-docgen` and realized that I was re-implementing a lot of the functionality in the core binary – so figured it would be worth while to see if making the binary into a Node stream is desirable.

- [x] **Create `src/parse-stream.js`:** a Node.js `Transform` stream in `objectMode`. If `{ paths }` are provided to a new `ParseStream` it begins the directory traversal in the next tick and emits Objects with this API:
``` js
{ file: '/path/to/component.js', ast: /* parsed AST */ }
```
- [x] **Ensure 100% non-breaking with existing CLI:** conforms to the existing CLI API, just refactoring under the covers so folks can do things like this and have strong guarantees of compatibility in the future. 
``` js
import { ParseStream } from 'react-docgen';
new ParseStream(/* opts */).on('readable', () => {});
```
- [ ] **Updates tests & docs:** not there yet (wants to get input first before wrapping it up), but if you folks are 👍on the change I'll fix all those up. Right now 7/14 the tests in `bin/__tests__` are passing.

### DIFF
--- a/bin/react-docgen.js
+++ b/bin/react-docgen.js
@@ -105,22 +105,25 @@ const parser = new ParseStream({
  * 1. No files passed, consume input stream
  */
 if (!Array.isArray(argv.args) || !argv.args.length) {
-  // var source = '';
-  // process.stdin.setEncoding('utf8');
-  // process.stdin.resume();
-  // var timer = setTimeout(function() {
-  //   process.stderr.write('Still waiting for std input...');
-  // }, 5000);
-  // process.stdin.on('data', function (chunk) {
-  //   clearTimeout(timer);
-  //   source += chunk;
-  // });
-  // process.stdin.on('end', function () {
-  //   try {
-  //     writeResult(parse(source));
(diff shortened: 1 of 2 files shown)


indexzero:
Bump @danez @fkling – would really appreciate your thoughts before continuing the refactor. Alternatively I could release the refactored code as a separate module and then consume it in this PR if that's preferable.

danez:
I guess if the cli keeps working as it is right now, abstracting more stuff out into reusable parts is totally fine.

Thanks for the effort 👍

indexzero:
Rad thanks for the input @danez – have a busy week next week, but after that I'll take another pass at this and get the tests cleaned up.

danez:
closing now, feel free to reopen if you ever consider to work on this again.

PR #269, opened 2018-05-24
Adding `@extends React.Component` support to isReactComponentClass

dustinsavery:
Certain react components may have multiple layers of inheritance and don't necessarily have a render function if it has a base class that handles rendering. In that case, `react-docgen` would fail to parse the file.

This PR:
- [x] Adds support for `@extends React.Component` in the class-level JSDoc block
- [x] Adds a unit test to verify functionality

### DIFF
--- a/src/utils/isReactComponentClass.js
+++ b/src/utils/isReactComponentClass.js
@@ -46,7 +46,7 @@ export default function isReactComponentClass(
 
   // check for @extends React.Component in docblock
   if (path.parentPath.value && path.parentPath.value) {
-    var classDeclaration;
+    var classDeclaration = {};
     if (Array.isArray(path.parentPath.value)) {
       var matches = path.parentPath.value.filter(function(declaration) { return declaration.type === 'ClassDeclaration' });
       if (matches[0]) {
--- a/src/utils/isReactComponentClass.js
+++ b/src/utils/isReactComponentClass.js
@@ -58,9 +58,8 @@ export default function isReactComponentClass(
     
     if (classDeclaration &&
       classDeclaration.leadingComments &&
-      classDeclaration.leadingComments.length &&
       classDeclaration.leadingComments.some(function (comment) {
-      return /@extends\s+React\.Component/.test(comment.value);
+        return /@extends\s+React\.Component/.test(comment.value);
       })) {
       return true;
     }
--- a/src/utils/isReactComponentClass.js
+++ b/src/utils/isReactComponentClass.js
@@ -44,6 +44,26 @@ export default function isReactComponentClass(
     return true;
   }
 
+  // check for @extends React.Component in docblock
+  if (path.parentPath.value && path.parentPath.value) {
+    var classDeclaration;
+    if (Array.isArray(path.parentPath.value)) {
+      var matches = path.parentPath.value.filter(function(declaration) { return declaration.type === 'ClassDeclaration' });
+      if (matches[0]) {
+        classDeclaration = matches[0];
+      }
+    } else {
+      classDeclaration = path.parentPath.value;
+    }
+    
+    if (classDeclaration.leadingComments && classDeclaration.leadingComments.length > 0) {
+      var matchedComments = classDeclaration.leadingComments.filter(function(comment) { return comment.value.match(/(@extends React.Component)/) });
+      if (matchedComments.length > 0) {
+        return true;
+      }
+    }
+  }
+
   // extends ReactComponent?
   if (!node.superClass) {
     return false;
--- a/src/utils/isReactComponentClass.js
+++ b/src/utils/isReactComponentClass.js
@@ -46,15 +46,9 @@ export default function isReactComponentClass(
 
   // check for @extends React.Component in docblock
   if (path.parentPath && path.parentPath.value) {
-    var classDeclaration;
-    if (Array.isArray(path.parentPath.value)) {
-      var matches = path.parentPath.value.filter(function(declaration) { return declaration.type === 'ClassDeclaration' });
-      if (matches[0]) {
-        classDeclaration = matches[0];
-      }
-    } else {
-      classDeclaration = path.parentPath.value;
-    }
+    var classDeclaration = Array.isArray(path.parentPath.value)
+      ? path.parentPath.value.find(function(declaration) { return declaration.type === 'ClassDeclaration' })
+      : path.parentPath.value;
     
     if (classDeclaration &&
       classDeclaration.leadingComments &&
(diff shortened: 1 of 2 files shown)


src/utils/isReactComponentClass.js
```diff
@@ -44,6 +44,26 @@ export default function isReactComponentClass(
     return true;
   }
 
+  // check for @extends React.Component in docblock
+  if (path.parentPath.value && path.parentPath.value) {
```
decompil3d: Why the doubled-up check?

src/utils/isReactComponentClass.js
```diff
@@ -44,6 +44,26 @@ export default function isReactComponentClass(
...
+    if (Array.isArray(path.parentPath.value)) {
+      var matches = path.parentPath.value.filter(function(declaration) { return declaration.type === 'ClassDeclaration' });
+      if (matches[0]) {
+        classDeclaration = matches[0];
+      }
+    } else {
+      classDeclaration = path.parentPath.value;
+    }
+    
+    if (classDeclaration.leadingComments && classDeclaration.leadingComments.length > 0) {
```
decompil3d: Be more defensive and add a check for whether `classDeclaration` is truthy:

```js
if (classDeclaration && classDeclaration.leadingComments && classDeclaration.leadingComments.length) {
```

(Note that checking the truthiness of `length` is sufficient rather than doing the explicit `> 0`)

src/utils/isReactComponentClass.js
```diff
@@ -44,6 +44,26 @@ export default function isReactComponentClass(
...
+      var matchedComments = classDeclaration.leadingComments.filter(function(comment) { return comment.value.match(/(@extends React.Component)/) });
```
decompil3d: A few issues:

1. No need to wrap the inside of the `RegExp` with parentheses, since you're not capturing anything
2. The `.` character in your `RegExp` will match anything -- you need to escape it to check for a literal period. Also, some people might indent the JSDoc field and value differently, so best to be lenient for that:

```js
/@extends\s+React\.Component/
```

3. No need to check through all of the comment lines -- you just need to find one that matches. So you can use `Array.prototype.some` to your advantage:

```js
if (classDeclaration.leadingComments.some(function (comment) {
  return /@extends\s+React\.Component/.test(comment.value);
})) {
  return true;
}
```

src/utils/isReactComponentClass.js
dustinsavery: Whoops, should've been `(path.parentPath&& path.parentPath.value)`

src/utils/isReactComponentClass.js
dustinsavery: Excellent points, getting all of that integrated.

src/utils/isReactComponentClass.js
```diff
@@ -56,11 +56,13 @@ export default function isReactComponentClass(
...
     }
     
-    if (classDeclaration.leadingComments && classDeclaration.leadingComments.length > 0) {
-      var matchedComments = classDeclaration.leadingComments.filter(function(comment) { return comment.value.match(/(@extends React.Component)/) });
-      if (matchedComments.length > 0) {
-        return true;
-      }
+    if (classDeclaration &&
+      classDeclaration.leadingComments &&
+      classDeclaration.leadingComments.length &&
```
decompil3d: No need for the length check, since `some` will handle an empty array just fine.

src/utils/isReactComponentClass.js
```diff
@@ -44,6 +44,28 @@ export default function isReactComponentClass(
...
+      }
+    } else {
+      classDeclaration = path.parentPath.value;
+    }
+    
+    if (classDeclaration &&
+      classDeclaration.leadingComments &&
+      classDeclaration.leadingComments.length &&
+      classDeclaration.leadingComments.some(function (comment) {
+      return /@extends\s+React\.Component/.test(comment.value);
```
decompil3d: Indent this line.

src/utils/isReactComponentClass.js
```diff
@@ -44,6 +44,28 @@ export default function isReactComponentClass(
...
+      if (matches[0]) {
+        classDeclaration = matches[0];
+      }
+    } else {
+      classDeclaration = path.parentPath.value;
+    }
+    
+    if (classDeclaration &&
+      classDeclaration.leadingComments &&
+      classDeclaration.leadingComments.length &&
```
decompil3d: No need for the length check as `some` will handle an empty array just fine.

src/utils/isReactComponentClass.js
```diff
@@ -44,6 +44,27 @@ export default function isReactComponentClass(
     return true;
   }
 
+  // check for @extends React.Component in docblock
+  if (path.parentPath && path.parentPath.value) {
+    var classDeclaration;
+    if (Array.isArray(path.parentPath.value)) {
+      var matches = path.parentPath.value.filter(function(declaration) { return declaration.type === 'ClassDeclaration' });
```
indexzero: Could use `.find` here fwiw, which would let you make this a one liner

``` js 
var classDeclaration = Array.isArray(path.parentPath.value)
  ?  path.parentPath.value.filter(function(declaration) { return declaration.type === 'ClassDeclaration' });
  : path.parentPath.value

File: src/utils/isReactComponentClass.js [hunk 8]
```diff
@@ -44,6 +44,27 @@ export default function isReactComponentClass(
...
+      if (matches[0]) {
+        classDeclaration = matches[0];
+      }
+    } else {
+      classDeclaration = path.parentPath.value;
+    }
+    
+    if (classDeclaration &&
```
indexzero: Could tighten this up a bit more:

``` js
if (!classDeclaration || !classDeclaration.leadingComments) {
  return false;
}

return classDeclaration.leadingComments.some(function (comment) {
  return /@extends\s+React\.Component/.test(comment.value);
});
```

src/utils/isReactComponentClass.js
dustinsavery: Nice! So much cleaner.

src/utils/isReactComponentClass.js
dustinsavery: I don't think we could do this. 
```
if (!classDeclaration || !classDeclaration.leadingComments) {
  return false;
}
```
^ This would cause an early return. We still need to check to see if the component directly extends from `React.Component`.

dustinsavery:
Ping @danez @fkling. Would appreciate your help reviewing this feature so that we can resume using the mainline of `react-docgen`.

PR #272, opened 2018-06-06
 Migrate babylon to @babel/parser

danez:
Add option to switch back to legacy decorators parsing

This is a breaking change as it will by default only parse the latest decorators spec which changed the syntax.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 `react-docgen` is a CLI and toolbox to help extracting information from [React][] components, and generate documentation from it.
 
-It uses [recast][] and [babylon][] to parse the source into an AST and provides methods to process this AST to extract the desired information. The output / return value is a JSON blob / JavaScript object.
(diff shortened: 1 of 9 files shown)


oktak:
Hello, 
Will this migration would support dynamic import() ?

Issue:
I've got **`SyntaxError: 'import' and 'export' may only appear at the top level`** when I use `import()` within the React Component class definition.

I've seen that the error is raised from `babylon\lib\parser\statement.js`
And I see that there is similar error message in babel-parser too:
https://github.com/babel/babel/blob/master/packages/babel-parser/src/parser/statement.js#L151

I would like to ask if React-docgen could add the options for Parser that `allowImportExportEverywhere` could set to the value that would eliminate the SyntaxError mentioned above.

Thank you!

Version and Platform:
- react-docgen: 2.21.0 `(npm view react-docgen version)`
- node v8.9.4
- npm v5.6.0
- windows 7 64bit

PR #273, opened 2018-06-07
Update normalizeClassDefinition.js

jquense:
Should i submit this against v3-dev? I meant too originally


jquense:
I wrote the test in a way the works in v1. I realized that i've not noticed this before because it seems like bablyon has only recently started making these ClassDeclarations, in the past they were expressions

danez:
Yeah it probably does not happen in v2 but does happen in the v3 beta. But good to have the check anyway.

Thank you very much.

PR #281, opened 2018-07-14
fix: correctly read description of inner types in arrayOf/objectOf

### LINKED ISSUE #152: PropTypes directly nested in arrayOf or ObjectOf
I've used react-docgen to create a [webpack plugin](https://github.com/OriR/react-docgen-plugin) that generates a markdown documentation. But I've encountered a small issue when trying to access the description of PropTypes directly nested inside `arrayOf` or `objectOf`.

In some of my components I'd like to describe the meaning of the type in the `arrayOf` or `objectOf` and not just the inner properties that each value needs to have. Currently I can achieve this in the description of `arrayOf` or `objectOf`, but this can look odd when a single object is described within the containing collection.

Given the example code below:
```javascript
/**
 * General component description.
 */
var Component = React.createClass({
  displayName: 'Component',

  propTypes: {
    /**
     * Comments of PropTypes that are direct children of arrayOf don't work.
     */
    prop: React.PropTypes.arrayOf(
      /**
       * This comment won't be generated in the resulting react-docgen object.
       */
      React.PropTypes.shape({
      /**
       * Simple Prop inside a nested shape works
       */
      subProp: React.PropTypes.number
    })),
    /**
     * This applies to objectOf as well.
     */
    anotherProp: React.PropTypes.objectOf(
      /**
       * Again, this comment won't be generated in the resulting react-docgen object.
       */
       React.PropTypes.number
    ).isRequired,
    /**
     * Top level shape works as expected.
     */
    topLevelShape: React.PropTypes.shape({
      /**
       * Index has a description
       */
      index: React.PropTypes.number,
      /**
       * Value has a description
       */
      value: React.PropTypes.string
    })
  },

  render: function() {
    // ...
  }
});
```

The resulting react-docgen object looks like this
```javascript
{
  "description": "General component description.",
  "displayName": "Component",
  "methods": [],
  "props": {
    "prop": {
      "type": {
        "name": "arrayOf",
        "value": {
          "name": "shape",
          "value": {
            "subProp": {
              "name": "number",
              "description": "Simple Prop inside a nested shape works",
              "required": false
            }
          }
        }
      },
      "required": false,
      "description": "Comments of PropTypes that are direct children of arrayOf don't work."
    },
    "anotherProp": {
      "type": {
        "name": "objectOf",
        "value": {
          "name": "number"
        }
      },
      "required": true,
      "description": "This applies to objectOf as well."
    },
    "topLevelShape": {
      "type": {
        "name": "shape",
        "value": {
          "index": {
            "name": "number",
            "description": "Index has a description",
            "required": false
          },
          "value": {
            "name": "string",
            "description": "Value has a description",
            "required": false
          }
        }
      },
      "required": false,
      "description": "Top level shape works as expected."
    }
  }
}
```
The comments above the directly nested types are not present.

danez:
Fixes #152

### DIFF
--- a/src/utils/__tests__/getPropType-test.js
+++ b/src/utils/__tests__/getPropType-test.js
@@ -266,6 +266,44 @@ describe('getPropType', () => {
     });
   });
 
+  it('detects descriptions on nested types in arrayOf', () => {
+    expect(
+      getPropType(
+        expression(`arrayOf(
+      /**
+       * test2
+       */
+      string
+    )`),
+      ),
+    ).toEqual({
+      name: 'arrayOf',
+      description: 'test2',
+      value: {
+        name: 'string',
+      },
+    });
+  });
+
+  it('detects descriptions on nested types in objectOf', () => {
+    expect(
+      getPropType(
+        expression(`objectOf(
+      /**
+       * test2
+       */
+      string
+    )`),
+      ),
+    ).toEqual({
+      name: 'objectOf',
+      description: 'test2',
+      value: {
+        name: 'string',
+      },
+    });
+  });
+
   it('detects descriptions on nested types in shapes', () => {
     expect(
       getPropType(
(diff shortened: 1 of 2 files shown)


greypants:
Woo! Just tried out RC3 and it works great. Thanks!

PR #290, opened 2018-08-22
Resolve flow $Keys to union when typeParameter is an ObjectTypeAnnotation

aracarie:
This adds support for ObjectTypeAnnotations as the typeParameter of $Keys.

This will now resolve to a union:

```
$Keys<{| a: string, b: string |}>
```

### DIFF
--- a/src/utils/getFlowType.js
+++ b/src/utils/getFlowType.js
@@ -68,6 +68,8 @@ function handleKeysHelper(path: NodePath): ?FlowElementsType {
   let value = path.get('typeParameters', 'params', 0);
   if (types.TypeofTypeAnnotation.check(value.node)) {
     value = value.get('argument', 'id');
+  } else if (!types.ObjectTypeAnnotation.check(value.node)) {
+    value = value.get('id');
   }
   const resolvedPath = resolveToValue(value);
   if (
--- a/src/utils/getFlowType.js
+++ b/src/utils/getFlowType.js
@@ -90,9 +90,9 @@ function handleKeysHelper(path: NodePath): ?FlowElementsType {
     return {
       name: 'union',
       raw: printValue(path),
-      elements: resolvedPath.node.properties.map(prop => ({
+      elements: resolvedPath.get('properties').map(prop => ({
         name: 'literal',
-        value: prop.key.name,
+        value: getPropertyName(prop),
       })),
     };
   }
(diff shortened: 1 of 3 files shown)


aracarie:
Refactored it a bit to use existing util function `getPropertyName`

src/utils/getFlowType.js
```diff
@@ -82,6 +82,19 @@ function handleKeysHelper(path: NodePath): ?FlowElementsType {
...
+    resolvedPath.node.properties.length
+  ) {
+    return {
+      name: 'union',
+      raw: printValue(path),
+      elements: resolvedPath.get('properties').map(prop => ({
+        name: 'literal',
+        value: getPropertyName(prop),
+      })),
+    };
```
fkling: I wonder if we can consolidate this even more since I guess nodes of type `ObjectExpression` and `ObjectTypeAnnotation` are very similar. At the very least, can we extract an array of key names for them and process that?

fkling:
Thank you for working on this! 😃

aracarie:
@fkling I updated `resolveObjectExpressionToNameArray` in resolveObjectKeysToArray, so that it handles ObjectExpressions and ObjectTypeAnnotations.

I also noticed that the handleKeysHelper was resolving this to a union:
```
const obj = { a: 'a', b: 'b' }
const test: $Keys<obj> = 'a'
```

Which I don't believe is correct. See this example on the [flow online repl](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBjOA7AzgFzDgCMArMAXjAG8wBDALjAHI7mAaMYp545sAL6YcBMPgCmBJgBIA0uICeuADwlSAPkos2QA). I removed this functionality and the associated test.

aracarie:
Any other changes needed here?

danez:
> I also noticed that the handleKeysHelper was resolving this to a union:
> 
> ```
> const obj = { a: 'a', b: 'b' }
> const test: $Keys<obj> = 'a'
> ```
> Which I don't believe is correct. See this example on the [flow online repl](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBjOA7AzgFzDgCMArMAXjAG8wBDALjAHI7mAaMYp545sAL6YcBMPgCmBJgBIA0uICeuADwlSAPkos2QA). I removed this functionality and the associated test.

Sorry for the late reply. Thanks for working on this.

I think we should leave the resolving of simple Identifiers in, as for example this could work:

```js
const obj = { a: 'a', b: 'b' }
type A = typeof obj;
const test: $Keys<A> = 'a'
```
This should resolve correctly to the same result as if doing an inline ObjectExpression, but currently returns unknown. But this is an other problem we can fix separately, but let's leave the resolving of `$Keys<A>` in for now and it's test. Other than that looks good to me, ready to merge.

aracarie:
@danez for sure, just pushed those changes.

danez:
Thank you very much.

PR #296, opened 2018-09-06
Creates the component tag handler

lucantini:
Hi! I've created a custom handler to add tags to Components.

The usage is simple, for example, a component with:
```
/**
* @beta
* @status new
* @subtitle This is a Button Component!
*/
class Button extends Component {
    ...
}
```

This should output:
```
{
    "displayName": "Button",
    "beta": true,
    "status": "new",
    "subtitle": "This is a Button Component!",
    ...
```

I needed this in my particular custom generated documentation! 

I can write tests and anything that is deemed necessary if you think this should be merged at all ;)

### DIFF
--- a/src/handlers/componentTagHandler.js
+++ b/src/handlers/componentTagHandler.js
@@ -1,0 +1,18 @@
+/* @flow */
+import type Documentation from '../Documentation';
+
+export default function displayNameHandler(documentation: Documentation) {
+  const description = documentation.get('description');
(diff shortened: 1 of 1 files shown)


danez:
Thanks for your PR.
In my last job we did exactly the same and did tag our components. :)

I think it is a really nice feature, but I wouldn't include it in the core of react-docgen. You could create a separate repository and publish it to npm similar to for example: https://github.com/siddharthkp/react-docgen-external-proptypes-handler.

PR #301, opened 2018-10-08
Allow descriptions for oneOfType values

### LINKED ISSUE #300: Support descriptions for oneOfType values
Hi there 👋 

I was digging through the code looking for ways to annotate propType values and found out that (luckily!) there's already support for that. As far as I can see, propType values can be annotated for values of:

- [`arrayOf`](https://github.com/reactjs/react-docgen/blob/master/src/utils/getPropType.js#L91)
- [`objectOf`](https://github.com/reactjs/react-docgen/blob/master/src/utils/getPropType.js#L110)
- [`shapeOf`](https://github.com/reactjs/react-docgen/blob/master/src/utils/getPropType.js#L143)

What do you think about adding support for `oneOfType` to that list? I'd also like to work on that if there's not a strong argument against it 🙂

cangoektas:
Hey!

Hope you don't mind me putting this PR together without a clear consent for this feature. I just had a lot of fun diving into this 😄 

### Changes
Allows values of `oneOfType` to have descriptions as well, similar to properties of `shape`. Fixes #300.

### Example

```js
{
  foo: PropTypes.shape({
    /** This was possible previously */
    a: PropTypes.string
  }),
  bar: PropTypes.oneOfType([
    /** Now this is possible, too */
    PropTypes.string,
    PropTypes.number
  ])
}
```

### DIFF
--- a/src/__tests__/__snapshots__/main-test.js.snap
+++ b/src/__tests__/__snapshots__/main-test.js.snap
@@ -889,3 +889,47 @@ Object {
   },
 }
 `;
+
+exports[`main fixtures processes component "component_17.js" without errors 1`] = `
+Object {
+  "description": "",
+  "displayName": "Foo",
+  "methods": Array [],
+  "props": Object {
+    "oneOfTypeProp": Object {
+      "description": "",
+      "required": false,
+      "type": Object {
+        "name": "union",
+        "value": Array [
+          Object {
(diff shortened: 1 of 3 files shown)


cangoektas:
@danez nice, thanks a lot! Is there also a chance to get this released soon? 😄

PR #311, opened 2018-11-26
Add forwardRef as a valid component definition

eps1lon:
- recognizes `React.forwardRef` as a valid component in a hoc
- adds support for `propTypes` and `defaultProps` for `React.forwardRef` components

Started out as a simple fix to recognize `export default hoc(React.forwardRef(...))` until I realized that `forwardRef` is not supported in `propTypes` handler.

I can split those two up if you want to (moving propTypes to fixture 14 and simply using 18 as the hoc+forwardRef fixture). As far as I understood the codebase this needs to be ported to every resolver? I haven't used flow in over a year so everything flow related (as most of the other additions) is just copy and paste.

### DIFF
--- a/src/__tests__/__snapshots__/main-test.js.snap
+++ b/src/__tests__/__snapshots__/main-test.js.snap
@@ -933,3 +933,33 @@ Object {
   },
 }
 `;
+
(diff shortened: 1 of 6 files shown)


danez:
Thank you.

PR #318, opened 2019-01-09
feat(resolve): resolve Object.values() in PropType.oneOf()

gatsbimantico:
Adding compatibility for Object.values as suggested on #211 

Creates a new util to extract keys from Object.values() and ObjectExpressions when used inside of oneOf()
It also follows spreads and tries to resolve them. Currently it bails out if either a Property is computed and the computed key is not a Literal or if a spread cannot be resolved.

Getters and setters will be ignored and undefined values transform to a null Literal.

Added the tests and checked that passes the `test:ci` script.

### DIFF
--- a/src/utils/__tests__/__snapshots__/getPropType-test.js.snap
+++ b/src/utils/__tests__/__snapshots__/getPropType-test.js.snap
@@ -134,7 +134,15 @@ Object {
 }
 `;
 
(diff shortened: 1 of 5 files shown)


gatsbimantico:
@danez Seems like VSCode and Travis are disagreeing on some prettier rules

PR #322, opened 2019-01-31
Create react-docgen playground website

danez:
This creates a mini playground for react-docgen, because I was sick of opening IDE, creating local testcases and running tests to reproduce bugs :D Can be improved if someone has the time, but it already serves its purpose for now.

Localy it can be run with `yarn start`

screen shot 2019-01-31 at 1 07 10 am

@fkling Let me know what you think and if you have any suggestions or legal requirements? If it is okay it needs github pages being turned on for the branch `gh-pages`. Could you do that? I guess the url would be https://reactcommunity.org/react-docgen/ Thanks.


website/. browserslistrc
```diff
@@ -0,0 +1,7 @@
+# Browsers that we support
```
fkling: Does the file name have an erroneous space between the `.` and the rest of the name?

fkling:
This is amazing! I love it ❤️ 

Should I turn on GitHub pages for the master branch?

fkling:
OK, I read https://docs.travis-ci.com/user/deployment/pages/ 😉

fkling:
Looks like we need to create an new gh-pages branch first, otherwise I cannot enable it for that branch.

Also, regarding the personal access token: It doesn't look like you can restrict a token to a specific repository, and I don't really want to create one for my personal account and add it to a travis config that can be accessed by other people.

Is the only solution to create a new user?

website/. browserslistrc
danez: upppss

danez:
Travis will create the branch on the first deployment I think. 
I already added the token to travis.

Let me fix that file. I can merge afterwards.

danez:
I created the gh-pages branch myself so it should be able to activate. The deployment does not yet work. Seems the changes to master since I branched for this PR make the build fail.
I will investigate.

fkling:
Ah I didn't see the token. Makes sense! Looks like gh-pages is activated in the settings, but http://reactcommunity.org/react-docgen doesn't work.

danez:
Maybe it takes some time?

Does it look like that? This is for react-tabs which is setup exactly like here.

screen shot 2019-02-07 at 12 12 54 am

danez:
Ah I think I remember. It only gets published once there is a new push event on the branch, let my quickly commit something.

danez:
Hmm not sure why it isn't published. Maybe this could help: https://webapps.stackexchange.com/a/104559 Basically it suggest to switch to master and then back to gh-pages.

fkling:
Switching to master and back to gh-pages seems to have worked!

PR #324, opened 2019-02-07
feat: support forwardRef for all resolvers

jquense:
extends #311 to all the resolvers


jquense:
what do you think about putting out a release with this? I know there is the babel config stuff to work around :/

danez:
I can do it tomorrow.

danez:
Okay I released it a 4.0.0.

PR #328, opened 2019-02-18
Fix docs on resolvers; add where to find them

maciej-gurban:
- Fixes markdown issue which was clipping the content of "resolvers" row
- Adds information where to find the resolvers

Before:
screen shot 2019-02-18 at 15 19 41

After:
screen shot 2019-02-18 at 15 23 13


fkling:
Thank you!

I would slightly tweak the statement to something like

> *Built-in* resolvers are available....

The function can (and should) be used with custom resolvers. I don't want to create the impression that only resolvers shipped with react-docgen can be used.

PR #330, opened 2019-02-20
feat: support PropTypes.elementType

eps1lon:
Introduced in facebook/prop-types#211, released in [`prop-types@15.7.0`](https://github.com/facebook/prop-types/blob/v15.7.0/CHANGELOG.md#1570).


danez:
Thanks.

PR #334, opened 2019-03-01
feat: support all the config resolving related babel options

jquense:
I realized trying to get this to work with gatsby in a monorepo that react-docgen doesn't have enough info to resolve my local configs and babelrc

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,6 @@ yarn add react-docgen --dev
 npm install --save-dev react-docgen
 ```
 
-
 ## CLI
 
 Installing the module adds a `react-docgen` executable which allows you to convert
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -362,18 +359,18 @@ we are getting this output:
...
+| Tuple                                     | `let x: [foo, "value", number];`                                                                                                                                        | `{ "name": "tuple", "raw": "<raw-signature>", "elements": [{ <element-type> }, ...] }`                                                                                                                                                                             |
+| Union                                     | `let x: number | string;`                                                                                                                                               | `{ "name": "union", "raw": "<raw-signature>", "elements": [{ <element-type> }, ...] }`                                                                                                                                                                             |
```
danez: Can we escape the pipe in the union somehow or use an escape sequence?

It is currently treated as table separator.

danez:
maybe this will also help @fkling with this problem https://github.com/reactjs/react-docgen/commit/60196fc30f6ed9ff79e0c241edc10ba4fc4758cc#comments?

README.md
jquense: bah sorry not sure what's happening thing, maybe prettier being over zealous...i'll fix

jquense:
ok readme fixed 👍

fkling:
So, I figured out what the issue is: The CI environment has a babel config file in its root that gets picked up. However, the necessary packages are not installed (and I can't get them to).

I'd like to disable loading the babel config file from the file system and I thought `babelrc: false` would do that, but it doesn't seem to work? Am I missing something here?

danez:
> Project-wide configs can also be disabled by setting "configFile" to false.

If you use `babel.config.js` this applies.

https://babeljs.io/docs/en/config-files

fkling:
Thanks, I have figured that out now as well 🤣  So many options...

fkling:
Not sure if ignoring local config files is a common enough issue to have a command line flag for...

PR #340, opened 2019-04-11
Handle computed properties correctly and do not fail generation

### LINKED ISSUE #317: dynamic context type names fail the whole file
3.X

dynamic context type names fail the whole file 
```
static contextTypes = {
    [DYNAMIC_NAME]: ...
}
```

```
TypeError: Property name must be an Identifier or a Literal

```

why fail the entire file?
a warning + moving to the next prop/context type should be the way to go.

thanks.

danez:
Fixes #317 

This ignores complex computed properties.
Non complex ones (identifiers, literals) are prefixed with `@computed#` and otherwise handled like normal.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -27,5 +27,9 @@ module.exports = {
         'no-unused-vars': 'off',
       },
     },
+    {
+      files: 'src/**/__tests__/*-test.js',
+      env: { jest: true },
+    },
   ],
 };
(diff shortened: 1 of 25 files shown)


fkling:
I didn't look at all the code, so maybe you already considered this: Should we try to resolve identifiers to their values and use that instead if it is a string or number?

danez:
That's a good idea, will add this.

rotawebmail:
dynamic context type names fail the whole file

static contextTypes = {
    [DYNAMIC_NAME]: ...
}
TypeError: Property name must be an Identifier or a Literal

why fail the entire file?
how to fix the problem?

PR #343, opened 2019-04-13
Assume HOC first param is Component not last

### LINKED ISSUE #342: React.memo not wonking
Code example:

```js
import React from "react";
import PropTypes from 'prop-types';

const Component = props => (....);

Component.propTypes = {...}
Component.defaultProps = {...}

export default React.memo(Component, (prevProps, nextProps) =>
  shallowCompare(prevProps, nextProps, 'data')
));
```

error with  `export default hoc(Component, params)`

danez:
Unless we know for sure it is not a component (literal, array, object, spread)

I did some research too and it seem most HOC have the component as first like for example `relay` and `React.memo`.

In case though someone wrote a HOC that assumes the component is the last argument, this will break their documentation.

Fixes #342

### DIFF
--- a/src/utils/__tests__/resolveHOC-test.js
+++ b/src/utils/__tests__/resolveHOC-test.js
@@ -19,33 +19,58 @@ describe('resolveHOC', () => {
   }
 
   it('resolves simple hoc', () => {
-    const path = parse(['hoc(42);'].join('\n'));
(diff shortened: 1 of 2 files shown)


Evalon:
This is great, thank you for your work! Will it be released soon?

kivervinicius:
?

PR #344, opened 2019-04-13
Detect components wrapped in HOC with all resolver

### LINKED ISSUE #331: Potential issue with findAllComponentDefinitions
We have a component that looks something like this:

```jsx
/**
 * @flow
 */
import * as React from 'react';

type Props = $ReadOnly<{|
  tabs: $ReadOnlyArray<string>,
|}>;

const TetraAdminTabs = React.memo<Props>((props: Props) => (
  <div></div>
));

export default TetraAdminTabs;
```

When I run `react-docgen` normally I get the information for the component. If I use `--resolver findAllComponentDefinitions`. I get `Error: No suitable component definition found`.

I think `findAllComponentDefinitions` should be able to find the default exported one.

----

I'll investigate this myself but just wanted to post to keep track of it.

danez:
Fixes #331

This is I think breaking as the following example will now detect a component:

```js
myFunctionCallWhichIsNotAHOC(something, node => (<div>{node}</div>), other);
```

And I don't see a way to workaround this as we cannot detect if a call is a HOC or not.

### DIFF
--- a/src/resolver/__tests__/findAllComponentDefinitions-test.js
+++ b/src/resolver/__tests__/findAllComponentDefinitions-test.js
@@ -227,4 +227,30 @@ describe('findAllComponentDefinitions', () => {
       expect(result[0].value.type).toEqual('CallExpression');
     });
   });
+
+  describe('regressions', () => {
+    it('finds component wrapped in HOC', () => {
+      const source = `
+        /**
+         * @flow
+         */
+        import * as React from 'react';
+
+        type Props = $ReadOnly<{|
+          tabs: $ReadOnlyArray<string>,
+        |}>;
+
+        const TetraAdminTabs = React.memo<Props>((props: Props) => (
+          <div></div>
+        ));
+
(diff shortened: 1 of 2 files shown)


fkling:
This should be possible to resolve without breaking changes. I think I have some local changes that rewrote `isReactForwardRefCall` to a more general `isReactHOC` call that also handles `React.memo`. Maybe we need to consolidate the different resolvers better.

However, I also think relaxing the rules around HOCs would be useful, but that should probably be in `resolveHOC`.

danez:
`React.forwardRef` though is not a HOC as far as I could tell. I did some simple tests in codesandbox:

So for HOC this works:

```js
Component = ...,

Component.propTypes = {};

export default HOC(Component);
```
This though does not work for `React.forwardRef`. 

There the static properties have to be assigned after the call:

```js
Component = React.forwardRef(...),

Component.propTypes = {};

export default Component;
```

PR #348, opened 2019-04-22
TypeScript support

### LINKED ISSUE #325: TypeScript support
Let's get `react-docgen` working with TypeScript. What will it take? 
`recast` added support for TypeScript in 2018, and from #316 it looks like @jquense has been working on getting the parser integrated with `react-docgen`.

I haven't worked with recast, and I'm not sure how different the output is between Flow and TS. 
But the the actual types are very similar. Here's the example from the Readme in TypeScript.
The only real change i did was changing `subvalue: ?boolean` to `subvalue: boolean | null`, which seems like the way to make values nullable.

## Input 

```typescript
import React, { Component } from 'react';

type Props = {
  /** Description of prop "foo". */
  primitive: number
  /** Description of prop "bar". */
  literalsAndUnion: 'string' | 'otherstring' | number
  arr: Array<any>
  func?: (value: string) => void
  noParameterName?: string => void
  obj?: { subvalue: boolean | null }
};

/**
 * General component description.
 */
export default class MyComponent extends Component<Props> {
  render() {
    // ...
  }
}
```

## Output

All flow types are currently exported under their own key "flowType". Following this structure, it would seem obvious to just add a new "typeScript" key?

```json
{
  "description":"General component description.",
  "props":{
    "primitive":{
      "typeScript":{ "name":"number" },
      "required":true,
      "description":"Description of prop \"foo\"."
    },
    "literalsAndUnion":{
      "typeScript":{
        "name":"union",
        "raw":"'string' | 'otherstring' | number",
        "elements":[
          { "name":"literal", "value":"'string'" },
          { "name":"literal", "value":"'otherstring'" },
          { "name":"number" }
        ]
      },
      "required":true,
      "description":"Description of prop \"bar\"."
    },
    "arr":{
      "typeScript":{
        "name":"Array",
        "elements":[
          { "name":"any" }
        ],
        "raw":"Array<any>"
      },
      "required":true
    },
    "func":{
      "typeScript":{
        "name":"signature",
        "type":"function",
        "raw":"(value: string) => void",
        "signature":{
          "arguments":[
            { "name":"value", "type":{ "name":"string" } }
          ],
          "return":{ "name":"void" }
        }
      },
      "required":false
    },
    "noParameterName":{
      "typeScript":{
        "name":"signature",
        "type":"function",
        "raw":"string => void",
        "signature":{
          "arguments":[
            { "name":"", "type":{ "name":"string" } }
          ],
          "return":{ "name":"void" }
        }
      },
      "required":false
    },
    "obj":{
      "typeScript":{
        "name":"signature",
        "type":"object",
        "raw":"{ subvalue: ?boolean }",
        "signature":{
          "properties":[
            {
              "key":"subvalue",
              "value":{
                "name":"boolean",
                "nullable":true,
                "required":true
              }
            }
          ]
        }
      },
      "required":false
    }
  }
}
```

devongovett:
Closes #325.
Depends on https://github.com/benjamn/ast-types/pull/330

This adds TypeScript support to react-docgen. It automatically replaces the flow plugin with the typescript plugin in babel-parser when a `.ts` or `.tsx` file is seen.

The output follows the existing flow output format since TypeScript types are quite similar to Flow. The implementation follows the existing flow implementation quite closely as well - a new `getTSType` file is added similar to the `getFlowType` one, and the surrounding code is updated to also handle TypeScript annotations in addition to Flow.

This brings up several questions about API and code organization.

1. Do we want to change the output format key from `flowType` to something more generic, add an additional `tsType` key, or just leave it as is?
2. Internally, do we want to leave the files/function names that now handle both flow and typescript with flow in the name, or change them to be more generic as well?

A few things left to do on this:

- [x] Support interfaces in addition to type aliases. This is also not currently the case for Flow, however...
- [x] Ensure that all TypeScript types are handled
- [x] Handle method documentation for TypeScript
- [x] Decide on questions above

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "commander": "^2.19.0",
     "doctrine": "^3.0.0",
     "node-dir": "^0.1.10",
-    "recast": "^0.17.3"
+    "recast": "^0.17.6"
   },
   "devDependencies": {
     "@babel/cli": "^7.0.0",
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
   "author": "Felix Kling",
   "license": "MIT",
   "dependencies": {
-    "@babel/core": "^7.0.0",
+    "@babel/core": "^7.4.4",
     "@babel/runtime": "^7.0.0",
     "async": "^2.1.4",
     "commander": "^2.19.0",
@@ -47,7 +47,6 @@
   },
   "devDependencies": {
     "@babel/cli": "^7.0.0",
-    "@babel/core": "^7.0.0",
     "@babel/plugin-proposal-object-rest-spread": "^7.0.0",
     "@babel/plugin-transform-runtime": "^7.0.0",
     "@babel/preset-env": "^7.0.0",
--- a/src/utils/getTSType.js
+++ b/src/utils/getTSType.js
@@ -1,0 +1,300 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+import getPropertyName from './getPropertyName';
+import printValue from './printValue';
+import recast from 'recast';
+import getTypeAnnotation from '../utils/getTypeAnnotation';
+import resolveToValue from '../utils/resolveToValue';
+import { resolveObjectToNameArray } from '../utils/resolveObjectKeysToArray';
+import type {
+  FlowTypeDescriptor,
+  FlowElementsType,
+  FlowFunctionSignatureType,
+  FlowObjectSignatureType,
+} from '../types';
+
+const {
+  types: { namedTypes: types },
+} = recast;
+
+const tsTypes = {
+  TSAnyKeyword: 'any',
+  TSBooleanKeyword: 'boolean',
+  TSUnknownKeyword: 'unknown',
+  TSNullKeyword: 'null',
+  TSUndefinedKeyword: 'undefined',
+  TSNumberKeyword: 'number',
+  TSStringKeyword: 'string',
+  TSSymbolKeyword: 'symbol',
+  TSThisType: 'this',
+  TSObjectKeyword: 'object',
+  TSVoidKeyword: 'void',
+};
+
+const namedTypes = {
+  TSArrayType: handleTSArrayType,
+  TSTypeReference: handleTSTypeReference,
+  TSTypeLiteral: handleTSTypeLiteral,
+  TSUnionType: handleTSUnionType,
+  TSFunctionType: handleTSFunctionType,
+  TSIntersectionType: handleTSIntersectionType,
+  TSTupleType: handleTSTupleType,
+  TSTypeQuery: handleTSTypeQuery,
+  TSTypeOperator: handleTSTypeOperator,
+};
+
+function handleTSArrayType(path: NodePath): FlowElementsType {
+  return {
+    name: 'Array',
+    elements: [getTSTypeWithResolvedTypes(path.get('elementType'))],
+    raw: printValue(path),
+  };
+}
+
+function handleTSTypeReference(path: NodePath): ?FlowTypeDescriptor {
+  let type: FlowTypeDescriptor;
+  if (types.TSQualifiedName.check(path.node.typeName)) {
+    const typeName = path.get('typeName');
+
+    if (typeName.node.left.name === 'React') {
+      type = {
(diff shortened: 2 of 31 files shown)


danez:
Thank you for working on this. This is amazing.

> 1. Do we want to change the output format key from flowType to something more generic, add an additional tsType key, or just leave it as is?

This is a tricky problem. I'm not sure about a clear answer. If we would add the docs for both type systems to one key we need to make sure they are always compatible and always stay compatible. Additional features which do not exist in the other language are okay I guess, but features that just would need a slightly different documentation could make things complicated. Maybe for example differences in enum types or mapped types? I can't think of any concrete example but I wonder if it would make it complicated for tools using react-docgen and supporting flow and ts.
Not sure if this is a real problem maybe you know more if there are any documentation differences in types between flow and ts?
If we would use different keys for ts and flow we wouldn't have this problem at all.

> 2. Internally, do we want to leave the files/function names that now handle both flow and typescript with flow in the name, or change them to be more generic as well?

I think we should rename it, but we don't have to do it in this PR. I would also classify this a breaking change, because I saw some custom resolvers using internal utils of react-docgen.

Also in addition what I said to 1. what do you think about adding a key to the documentation that indicates which language we were parsing, js, flow or typescript? But also not sure if this would be useful to users. 

Maybe @okonet or @sapegin have input about what would make it easy to use in react-styleguidist?

devongovett:
Thanks for your thoughts on this. I think we should have a separate `tsType` (or `typescript`?) key from `flowType` for future extensions, though the format will probably be the same as flow initially. This way it is also not a breaking change.

thebuilder:
I think it also makes the most sense to keep `flowType` as is, and introduce `tsType`. 
If you are consuming the output, then you already have the handle both `PropTypes` and `flowType`. Adding an extra key, that hopefully matches the structure of `flowType`, should be easy enough to deal with.

devongovett:
I think this is ready for review! I adjusted the output to use `tsType` for TypeScript and retain`flowType` for Flow. They use the same output format at the moment though. I didn't rename any of the internal files or methods for flow that now also handle TypeScript since @danez mentioned that would be a breaking change.

Also added support for generics to both Flow and TypeScript. It will inline type parameters when merging extended interfaces or other type references. This should help allow you to reuse generic types across components while keeping the docs understandable.

src/utils/getTypeParameters.js
```diff
@@ -0,0 +1,47 @@
...
+
+export type TypeParameters = {
+  [string]: NodePath,
+};
+
+export default function getTypeParameters(
+  declaration: NodePath,
+  instantiation: NodePath,
+  inputParams: ?TypeParameters,
+) {
```
danez: Could you add a return type here? Which I guess is `TypeParameters`?

src/utils/getTypeParameters.js
```diff
@@ -0,0 +1,47 @@
...
+  const params = {};
+
+  let i = 0;
+  declaration.get('params').each((paramPath: NodePath) => {
+    const key = paramPath.node.name;
+    const defaultTypePath = paramPath.node.default
+      ? paramPath.get('default')
+      : null;
+    const typePath =
+      i < instantiation.node.params.length
```
danez: micro optimizing: The whole length lookup could be done before the loop.

src/utils/getTSType.js
```diff
@@ -0,0 +1,383 @@
...
+    signature: {
+      arguments: [],
+      return: getTSTypeWithResolvedTypes(
+        path.get('typeAnnotation'),
+        typeParams,
+      ),
+    },
+  };
+
+  path.get('parameters').each(param => {
```
danez: Should we treat the [`this` parameter](https://www.typescriptlang.org/docs/handbook/functions.html#this-parameters) different here and maybe assign it to it's own property `type.signature.this`? Just wonder because it is not a real parameter.
It probably does not make that much of a difference, as consumers of react-docgen could simply do that themself, but imho makes it a little bit easier to use.
This though would be the first difference to flow though if I'm correct :D

package.json
```diff
@@ -43,7 +43,7 @@
     "commander": "^2.19.0",
     "doctrine": "^3.0.0",
```
danez: We should update `@babel/core` to `^7.4.4` so we ensure that typescript is fully supported.

Also noticed we have it in dependencies and dev dependencies. Having it only in dependencies is probably fine.

devongovett:
Thanks for the review @danez! Updated based on your comments.

devongovett:
So I upgraded babel-core, and am now getting a bunch of errors about undefined exports in the tests... :|

danez:
I will have a look

PR #349, opened 2019-04-23
Remove recast and use ast-types directly

danez:
This removes recast and only uses ast-types directly.

The performance increases significantly, but I haven't tested more different cases. 

There are still some cases that do not yet work:

- [x] Printing of nodes which were created in react-docgen by ast-types builders.
- [x] Comments in printed code are not filtered out

Fixes #237

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -16,6 +16,7 @@ module.exports = {
   globals: {
     ASTNode: true,
     NodePath: true,
+    $Exact: true,
   },
   overrides: [
     {
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -16,7 +16,6 @@ module.exports = {
   globals: {
     ASTNode: true,
     NodePath: true,
-    Recast: true,
   },
   overrides: [
     {
@@ -28,7 +27,7 @@ module.exports = {
       },
     },
     {
-      files: 'src/**/__tests__/*-test.js',
+      files: '@(src|bin)/**/__tests__/*-test.js',
       env: { jest: true },
     },
   ],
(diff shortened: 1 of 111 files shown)


fkling:
Amazing. Somehow I thought the whole `NodePath` business was part of recast (that would probably be more difficult to remove).

Looking forward to where this goes 😃 

> The performance increases significantly, but I haven't tested more different cases.

Can you share numbers?

danez:
yeah I forgot to include that yesterday:

Before:
```
❯ node --expose-gc --max-old-space-size=2000 --max-semi-space-size=1500 --noconcurrent_sweeping ./benchmark/index.js
Node: v10.15.1
Running benchmark for CircularProgress.js ...
┌─────────────────────┬─────────────────────────────┐
│ fixture             │ timing                      │
├─────────────────────┼─────────────────────────────┤
│ CircularProgress.js │ 30.34 ops/sec ±7.82% (33ms) │
└─────────────────────┴─────────────────────────────┘
```

With PR:
```
❯ node --expose-gc --max-old-space-size=2000 --max-semi-space-size=1500 --noconcurrent_sweeping ./benchmark/index.js
Node: v10.15.1
Running benchmark for CircularProgress.js ...
┌─────────────────────┬─────────────────────────────┐
│ fixture             │ timing                      │
├─────────────────────┼─────────────────────────────┤
│ CircularProgress.js │ 40.55 ops/sec ±5.08% (25ms) │
└─────────────────────┴─────────────────────────────┘
```

> The commandline options are to prevent gc to interfere, so the script runs without automatic gc.

The test file is taken from material ui. I need to find bigger or different (flow) files to test on. Would also be interesting to run this on a big codebase and see the difference.

fkling:
>  Would also be interesting to run this on a big codebase and see the difference.

I get the hint 😉

PR #350, opened 2019-04-24
feat(defaultPropsHandler): Fully support forwardRef

eps1lon:
In `React.forwardRef(({ foo = 'bar' }, ref) => <div ref={ref}>{foo}</div>)` the default props handler would not detect that `foo` had a default value.

I was curious if #343  would've fixed this as well but this is not the case.

### DIFF
--- a/src/handlers/defaultPropsHandler.js
+++ b/src/handlers/defaultPropsHandler.js
@@ -15,7 +15,8 @@ import printValue from '../utils/printValue';
 import recast from 'recast';
 import resolveToValue from '../utils/resolveToValue';
 import resolveFunctionDefinitionToReturnValue from '../utils/resolveFunctionDefinitionToReturnValue';
-import isStatelessComponent from '../utils/isStatelessComponent';
+import isReactComponentClass from '../utils/isReactComponentClass';
+import isReactForwardRefCall from '../utils/isReactForwardRefCall';
 
 const {
   types: { namedTypes: types },
@@ -53,7 +54,12 @@ function getDefaultValue(path: NodePath) {
 }
 
 function getStatelessPropsPath(componentDefinition): NodePath {
-  return resolveToValue(componentDefinition).get('params', 0);
+  const value = resolveToValue(componentDefinition);
+  if (isReactForwardRefCall(value)) {
+    const inner = value.get('arguments', 0);
+    return inner.get('params', 0);
+  }
+  return value.get('params', 0);
 }
 
 function getDefaultPropsPath(componentDefinition: NodePath): ?NodePath {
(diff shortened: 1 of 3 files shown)


src/handlers/defaultPropsHandler.js
```diff
@@ -118,7 +124,10 @@ export default function defaultPropsHandler(
 ) {
   let statelessProps = null;
   const defaultPropsPath = getDefaultPropsPath(componentDefinition);
-  if (isStatelessComponent(componentDefinition)) {
+  /**
+   * function, lazy, memo, forwardRef etc components can resolve default props as well
+   */
+  if (!isReactComponentClass(componentDefinition)) {
     statelessProps = getStatelessPropsPath(componentDefinition);
```
eps1lon: Happy to change that too. This was always a bit misleading and with hooks it is definitely wrong.

danez: What are you referring here? Changing the name?

danez:
Thanks for your contribution again. Looks good to me, I wonder can you confirm it also works when the Component is not directly nested?

Like:

```
import React from 'react';

const Component = ({ foo = 'bar' }, ref) => (<div ref={ref}>{foo}</div>);
React.forwardRef(Component);
```

By looking at the code we might have to do another `resolveValue`, but haven't tested it.

src/handlers/defaultPropsHandler.js
eps1lon: Naming yes. Doesn't matter really since it's internals only. This should probably be handled in a different PR and applied consistently. But I don't know if this would change some react-docgen/utils API so I guess this is fine.

eps1lon:
> Thanks for your contribution again. Looks good to me, I just wonder can you confirm it also works when the Component is not directly nested?
> 
> Like:
> 
> ```js
> import React from 'react';
> 
> const Component = ({ foo = 'bar' }, ref) => (<div ref={ref}>{foo}</div>);
> React.forwardRef(Component);
> ```
> 
> By looking at the code we might have to do another `resolveValue`, but haven't tested it.

I'm not sure we should support this pattern since the argument for forwardRef isn't a component but a render function that takes a second argument. This may be technically only but you couldn't use `(props, ref) => <div />` on its own. Or add `propTypes` to the render function e.g.
```js
const Component = (props, ref) => {}
Component.propTypes = {}; // Warning: forwardRef render functions do not support propTypes or defaultProps. Did you accidentally pass a React component?
export default React.forwardRef(Component);
```

If you think this is too rigid I can add an additional test and see if it works.

danez:
I see that makes sense. 👍

src/handlers/defaultPropsHandler.js
danez: What would you rename it to?

eps1lon: I don't know. Probably just getFunctionPropsPath or just getPropsPath. Naming is hard but I don't think this discussion serves any purpose.

PR #351, opened 2019-04-24
fix: Add missing isReactForwardRefCall export

eps1lon:
Forgot to export this in #311. Need this for a custom default props resolver and I'd rather use the public interface.


danez:
Thanks

jacobbogers:
Hi, I have some codes that uses `isReactForwardRefCall`  what version should i use?

Issue #361, opened 2019-05-18
TypeScript mapped type support


ahutchings:
Docgen is outputting parsing TypeScript mapped types as `{name: 'unknown'}` . I've started putting together a change to add support for mapped types, but I'm not sure how we want to represent the type descriptor for it, since it's not a feature that Flow has. Does anyone have thoughts on how the type descriptor should be structured for something like this?

```typescript
var x: MyMappedType = {top: true};

type Side = 'top' | 'right' | 'bottom' | 'left';
type MyMappedType = { [side in Side]?: boolean};
```

cc @devongovett

danez:
Isn't this the same as https://flow.org/en/docs/types/objects/#toc-objects-as-maps ?

ahutchings:
Yep, looks like it is, and we can use the existing FlowObjectSignatureType to represent it. PR incoming shortly.

PR #362, opened 2019-05-19
Add support for TSMappedType

ahutchings:
Refs #361


ahutchings:
Not sure if we should merge this - the name of the type parameter is currently lost in the translation to the object signature.

danez:
@devongovett Can you please have a look at this?

danez:
We can add the name of the key later if there is demand, but we also do not add it for flow atm.

Issue #366, opened 2019-05-21
RefTypes inside Interface does not expand correctly in Typescript


sylvesteraswin:
## Original Example (fixtures_23.tsx)
```tsx
/**
 * Copyright (c) Facebook, Inc. and its affiliates.
 *
 * This source code is licensed under the MIT license found in the
 * LICENSE file in the root directory of this source tree.
 *
 */

import React, { Component } from 'react';

type BaseProps = {
  /** Optional prop */
  foo?: string,
  /** Required prop */
  bar: number
};

type TransitionDuration = number | { enter?: number, exit?: number } | 'auto';

type Props = BaseProps & {
  /** Complex union prop */
  baz: TransitionDuration
}

/**
 * This is a typescript class component
 */
export default class TSComponent extends Component<Props> {
  render() {
    return <h1>Hello world</h1>;
  }
}
```
The output for the above example seems to be generated correctly.

## Modified example with interfaces

```tsx
import React, { Component } from "react";

interface BaseProps {
  /** Optional prop */
  foo?: string;
  /** Required prop */
  bar: number;
}

type TransitionDuration = number | { enter?: number; exit?: number } | "auto";

interface Props {
  something: BaseProps;
  /** Complex union prop */
  baz: TransitionDuration;
}

/**
 * This is a typescript class component
 */
export default class TSComponent extends Component<Props> {
  render() {
    return <h1>Hello world</h1>;
  }
}
```

#### Output
```json
{
	"description": "This is a typescript class component",
	"displayName": "TSComponent",
	"methods": [],
	"props": {
		"something": {
			"required": true,
			"tsType": {
				"name": "BaseProps"
			},
			"description": ""
		},
		"baz": {
			"required": true,
			"tsType": {
				"name": "union",
				"raw": "number | { enter?: number; exit?: number } | \"auto\"",
				"elements": [{
					"name": "number"
				}, {
					"name": "signature",
					"type": "object",
					"raw": "{ enter?: number; exit?: number }",
					"signature": {
						"properties": [{
							"key": "enter",
							"value": {
								"name": "number",
								"required": false
							}
						}, {
							"key": "exit",
							"value": {
								"name": "number",
								"required": false
							}
						}]
					}
				}, {
					"name": "literal",
					"value": "\"auto\""
				}]
			},
			"description": "Complex union prop"
		}
	}
}
```

#### Issue
`props.something` in the above output has only the `name` and not the `interface` which is defines.

[...]

PR #369, opened 2019-06-01
Added handleTSIndexedAccessType.

### LINKED ISSUE #367: Handle TSIndexedAccessType
There currently isn't a handler for `TSIndexedAccessType`:

```ts
type ABC = {
  foo: string
}

type DEF = {
  bar: ABC["foo"] // <-- indexed type
}
```
The output for bar is `{ name: "unknown" }`. Expected output would be either:

`{ name: "string" }` or;
`{ name: "ABC[\"foo\"]" }`

I believe there is an ongoing discussion in #366 over which of these 2 it should be (based on whether ABC is declared as a type or an interface), but either way currently it is neither.

I'm happy to attempt a PR but I have zero experience with AST so perhaps only if this doesn't get picked up by someone more capable 😃

thchia:
Fixes #367

I'm new to working with ASTs, and welcome any feedback. I also had some issues with type narrowing in Flow, resulting in the use of `$Shape<>`.

### DIFF
--- a/src/utils/getTSType.js
+++ b/src/utils/getTSType.js
@@ -19,6 +19,7 @@ import getTypeParameters, {
 import type {
   FlowElementsType,
   FlowFunctionArgumentType,
+  FlowLiteralType,
   FlowObjectSignatureType,
   FlowSimpleType,
   FlowTypeDescriptor,
@@ -53,6 +54,7 @@ const namedTypes = {
   TSTupleType: handleTSTupleType,
   TSTypeQuery: handleTSTypeQuery,
   TSTypeOperator: handleTSTypeOperator,
+  TSIndexedAccessType: handleTSIndexedAccessType,
 };
 
 function handleTSArrayType(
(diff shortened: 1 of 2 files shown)


src/utils/getTSType.js
```diff
@@ -318,6 +320,40 @@ function handleTSTypeOperator(path: NodePath): ?FlowTypeDescriptor {
...
+    path.get('objectType'),
+    typeParams,
+  );
+  // eslint-disable-next-line no-undef
+  const indexType: $Shape<FlowLiteralType> = getTSTypeWithResolvedTypes(
+    path.get('indexType'),
+    typeParams,
+  );
+
+  // We only get the signature if the objectType is a type (vs interface)
```
devongovett: Why not for interfaces?

thchia: @devongovett actually I'm not 100% sure why, but I believe in my trial and error when defining the objectType as an interface, there is no signature property on it. I'm not sure if it is related to the discussion [here](https://github.com/reactjs/react-docgen/issues/366#issuecomment-495715639)?

sylvesteraswin:
@thchia, thanks a lot for working on this. Whats the process on this PR?

thchia:
@sylvesteraswin at this stage I’m still waiting on any additional comments from @devongovett :)

danez:
I think we need to look at interfaces in a separate change. I think the change (for object types) makes sense.

PR #373, opened 2019-06-06
Updated example with valid isRequired

### LINKED ISSUE #372: Broken Example
The example at https://reactcommunity.org/react-docgen/ is returning **false** for  _foo: PropTypes.number.isRequired_

MikeLockz:
Resolves #372 by moving the isRequired PropType definition to a prop that does not have a defaultProp defined.


danez:
Thanks

PR #400, opened 2019-10-30
check for value before using methods

bryceosterhaus:
Not sure if this is the best way to fix this, but I was running into this error when using using typescript with this sort of type

```ts
const VALUES = ['one', 'two'];

interface IProps {
    values?: typeof VALUES[number]
}
```

Check out the failing fixture for an example


bryceosterhaus:
Any update on this?

bryceosterhaus:
Hey @danez, just curious if you think this might be able to make it into a release at any point. Not sure how active this project is, but wanted to follow up in case this PR got lost in the mix of notifications.

danez:
I do not know what happened to this PR. It must have fallen under the radar. I'm really sorry. It is merged now, I will backport this to 5.x and cut a release soon.

bryceosterhaus:
No worries! Thanks for merging

PR #408, opened 2019-12-11
fix(typescript): Support type as expressions and type assertions

### LINKED ISSUE #401: Support for TSAsExpression ast type?
I currently have the following code, and it breaks when I run parse with the error `No suitable component definition found`

```tsx
import React from 'react';

const Group = React.forwardRef<
	HTMLDivElement,
	React.HTMLAttributes<HTMLDivElement>
>((props, ref) => {
	return <div {...props} ref={ref} />;
});

type TInput = React.ForwardRefExoticComponent<
	React.HTMLAttributes<HTMLInputElement> &
		React.RefAttributes<HTMLInputElement>
> & {Group: typeof Group};

const Input = React.forwardRef<
	HTMLInputElement,
	React.HTMLAttributes<HTMLInputElement>
>((props, ref) => {
	return <input {...props} ref={ref} />;
}) as TInput;

Input.Group = Group;

export default Input;
```

danez:
Fixes #401

### DIFF
--- a/src/utils/__tests__/resolveToValue-test.js
+++ b/src/utils/__tests__/resolveToValue-test.js
@@ -11,8 +11,8 @@ import { parse } from '../../../tests/utils';
 import resolveToValue from '../resolveToValue';
 
 describe('resolveToValue', () => {
-  function parsePath(src) {
-    const root = parse(src.trim());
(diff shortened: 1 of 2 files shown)


bryceosterhaus:
Thank you!

PR #409, opened 2019-12-11
feat(typescript): add private modifier for private methods

satya164:
When generating documentation for methods in class components using TypeScript, I have some methods marked as `private` because they are internal and don't want to show them in the docs. But I don't have any information regarding whether a method is public or private in the currently extracted metadata. In this PR, I'm adding a modifier for private methods so I can know which methods to exclude.

### DIFF
--- a/src/__tests__/__snapshots__/main-test.js.snap
+++ b/src/__tests__/__snapshots__/main-test.js.snap
@@ -1398,7 +1398,47 @@ Object {
       "description": "This is a method",
       "docblock": "This is a method",
       "modifiers": Array [],
-      "name": "method",
+      "name": "foo",
+      "params": Array [
+        Object {
+          "name": "a",
+          "type": Object {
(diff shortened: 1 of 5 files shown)


danez:
Thanks. I was just thinking we probably can do the same for [private methods in javascript](https://github.com/tc39/proposal-private-methods) but what react-docgen is actually doing is documenting the API if a react component, so props and (public) methods. I even think we probably shouldn't even include private methods.
I guess that would work for you too?

satya164:
@danez I even think we probably shouldn't even include private methods. I guess that would work for you too?

It works for me. Should I update the PR to filter out private methods?

danez:
> 
> It works for me. Should I update the PR to filter out private methods?

yes please

danez:
Thank you, I made the change to skip private methods. It will be in the next version.

PR #412, opened 2019-12-21
Add markdown cli

younabobo:
Adds an alternative command-line utility that works the same way, except that it generates output in a markdown format and writes it to the current directory


danez:
The markdown generator is only an example. We do not want to support and maintain an markdown parser for react-docgen.

Feel free to create your own npm package that creates markdown files from the react-docgen output.

Issue #427, opened 2020-01-24
Flow Maximum Call stack exceeded on recursive type structures


FermiDirak:
Likely related to https://github.com/reactjs/react-docgen/issues/215. Recursive type structs cause a maximum call-stack exceeded error when parsing with Flow.

```jsx
import * as React from "react";

type Identity<T> = T;

type Props<T> = {
  prop: Identity<T>
};

export default function MyComponent<T>(props: Props<T>) {
  return <div/>
}
```

PR #428, opened 2020-01-24
Fix to Flow Maximum Callstack exceeded error on recursive generic structs

FermiDirak:
Resolves https://github.com/reactjs/react-docgen/issues/427

Nested structs using generics currently cause a `maximum callstack exceeded` error. This PR resolves the issue by short-circuiting the recursion when the type resolves to a generic.

This approach has a shortcoming where the below example will resolve the prop `prop` as `T` instead of as `string`. But this is currently the same behavior seen when switching to typescript and should be fine since it's an an overall improvement to docgen erroring.

```jsx
import * as React from "react";

type Identity<T> = T;

type Props<T> = {
  prop: Identity<T>;
}

export default function Test(props: Props<string>) {
  return <div />;
}
```

I'm happy to iterate on this PR, let me know if you need me to change anything.

### DIFF
--- a/src/__tests__/__snapshots__/main-test.js.snap
+++ b/src/__tests__/__snapshots__/main-test.js.snap
@@ -1516,3 +1516,64 @@ Object {
   },
 }
 `;
+
+exports[`main fixtures processes component "component_32.js" without errors 1`] = `
+Object {
+  "description": "",
+  "displayName": "Segments",
+  "methods": Array [
+    Object {
+      "docblock": null,
+      "modifiers": Array [],
+      "name": "foo",
+      "params": Array [
+        Object {
+          "name": "props",
+          "optional": undefined,
+          "type": Object {
+            "alias": "Props",
+            "name": "signature",
+            "raw": "{
+  segments: Array<T>,
+}",
+            "signature": Object {
+              "properties": Array [
+                Object {
+                  "key": "segments",
+                  "value": Object {
+                    "elements": Array [
+                      Object {
+                        "name": "T",
+                      },
+                    ],
+                    "name": "Array",
+                    "raw": "Array<T>",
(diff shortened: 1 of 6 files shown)


TheSavior:
We are running into a similar problem at Facebook on 5.1.0. I haven't gotten it down to a minimal repro yet, but our stack trace looks like this:
```
RangeError: Maximum call stack size exceeded
        at Object.hasOwnProperty (<anonymous>)

      at defFromValue (node_modules/ast-types/lib/types.js:339:24)
      at DefImpl.check (node_modules/ast-types/lib/types.js:362:24)
      at PredicateType.predicate (node_modules/ast-types/lib/types.js:351:109)
      at PredicateType.Object.<anonymous>.PredicateType.check (node_modules/ast-types/lib/types.js:122:27)
      at Object.handleGenericTypeAnnotation [as GenericTypeAnnotation] (node_modules/react-docgen/dist/utils/getFlowType.js:109:52)
      at getFlowTypeWithResolvedTypes (node_modules/react-docgen/dist/utils/getFlowType.js:296:33)
      at Object.handleGenericTypeAnnotation [as GenericTypeAnnotation] (node_modules/react-docgen/dist/utils/getFlowType.js:135:12)
      at getFlowTypeWithResolvedTypes (node_modules/react-docgen/dist/utils/getFlowType.js:296:33)
      at Object.handleGenericTypeAnnotation [as GenericTypeAnnotation] (node_modules/react-docgen/dist/utils/getFlowType.js:135:12)
```

Does this line up with the stack trace that this PR is fixing?

FermiDirak:
@TheSavior I've added the minimal repo you've listed, confirming the fix works for class components as well.

danez:
Thanks for the fix.

Issue #244, opened 2018-01-24
Better documentation on resolvers


acondiff:
I am trying to write a script to generate markdown. I have reviewed the documentation for `parse()` and I am trying to use the `findAllComponentDefinitions` resolver similar to `react-docgen ./myfile.js --resolver findAllComponentDefinitions -o ./docs` in the command line. This is what I attempted, but it doesnt work. 

```
const componentAPI = reactDocs.parse(data, reactDocs.findAllComponentDefinitions)
const markdown = generateMarkdown(components[i], componentAPI)
```

The current documentation does not seem very clear to me.

saoirse-zee:
Hi @acondiff -- I had the same issue, and had to dig around in the source code to find something that worked. 

The resolvers are exposed on a property called "resolver":
```
const reactDocs = require('react-docgen')
const resolvers = reactDocs.resolver
```

I think you can get your stuff working if you change your first line to this:
```
const componentAPI = reactDocs.parse(data, reactDocs.resolver.findAllComponentDefinitions)
```

maciej-gurban:
Thanks a bunch for this. Made a PR to update the docs with this info (and fix the broken description in the table): https://github.com/reactjs/react-docgen/compare/master...maciej-gurban:patch-1

PR #451, opened 2020-04-15
Improving resolver documentation 2.0

nahumzs:
I tried to understand how to use the resolvers and was not clear for me how to do that until googling and landing in this issue https://github.com/reactjs/react-docgen/issues/244 which asked to improve the documentation, which ended with this PR https://github.com/reactjs/react-docgen/pull/328 which improved as well the documentation, but wasn't that clear to me, because I read that part of the documentation, I think adding the `reactDocs.parse` is crystal clear how to use it.


danez:
Thanks

PR #454, opened 2020-06-02
Basic React Hooks support (useImperativeHandle only)

JaredReisinger:
Add detection of `useImperativeHandle()`-defined methods on components, including test cases and README documentation.

> _**Additional background:** I have a component I've just refactored from a class-based implementation to the new hooks-style implementation, and am publishing the documentation via Styleguidist.  Somewhat to my surprise, the once-public methods on the class, which are now exposed via `useImperativeHandle()`, are no longer showing up in the generated documentation.  Digging in, this appears to be the correct place to get those methods exposed._

Note that this does _**not**_ address the desire to document custom hooks, nor does it find `useCallback()` handlers as those aren't exposed in an imperative way and thus shouldn't be considered a part of a component's public API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,7 +368,63 @@ we are getting this output:
 }
 ```
 
-### Types
+## React Hooks support
+
+If you are using React Hooks, react-docgen will now also find component methods defined directly via the `useImperativeHandle()` hook.
+
+> **Note**: react-docgen will not be able to grab the type definition if the type is imported or declared in a different file.
+
+### Example
+
+For the following component using `useImperativeHandle`:
+
+
+```js
+import React, { useImperativeHandle } from 'react';
+
+/**
+ * General component description.
+ */
+const MyComponent = React.forwardRef((props, ref) => {
+
+  useImperativeHandle(ref, () => ({
+    /**
+     * This is my method
+     */
+    myMethod: (arg1) => {},
+  }));
+
+  return /* ... */;
+});
+
+export default MyComponent;
+```
+
+we are getting this output:
+
+```json
+{
+  "description": "General component description.",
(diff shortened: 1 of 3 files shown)


JaredReisinger:
@danez, has react-docgen gone into maintenance mode?  I know that for the most part, there's not a lot to do since Javascript hasn't changed that much.  That said, React added hooks in [16.8.0 (February 6, 2019)](https://github.com/facebook/react/blob/master/CHANGELOG.md#1680-february-6-2019), 19 months ago, so adding support for this isn't exactly "cutting edge" at this point.

I've tried to make sure the proposed change is well documented and tested... is there anything else I can do to make it easier or less-risky to merge?

README.md
```diff
@@ -368,7 +368,63 @@ we are getting this output:
 }
 ```
 
-### Types
+## React Hooks support
+
+If you are using React Hooks, react-docgen will now also find component methods defined directly via the `useImperativeHandle()` hook.
+
+> **Note**: react-docgen will not be able to grab the type definition if the type is imported or declared in a different file.
```
danez, line 375: Do you think this could work with the work that has been done in #464?

danez (approved):
Really sorry for the delay. It looks good to me, was just wondering about the note in the readme.

README.md
JaredReisinger: I don't know.... I'll take a look at #464 and try to understand what it's doing.  At first blush, it seems like yes, this might exactly solve that problem, I'll just need to understand what changes I'll need to make to use the new functionality (or whether things have changed in such a way that I don't have to do anything, and just get the new behavior for free 🤞 ).

PR #471, opened 2020-10-09
Version bump to fix dependency resolutions

ethancrook99:
Bumped version to 5.3.1 to fix dependencies that aren't resolving correctly. Specifically, `ast-types` is resolving to `^0.13.4` as the dependency for this project, when it should be resolving to `^0.14.2`. Going along with this PR, the new version would also need to be published to NPM.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-docgen",
-  "version": "5.3.0",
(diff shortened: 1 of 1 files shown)


danez:
Thanks but this will be done automatically on releasing

PR #592, opened 2022-05-12
Support 'memo(forwardRef(...'

mfazekas:
While `forwardRef(...)`  is already supported `memo(forwardRef(...))` doesn't.


danez:
Could you show me an example of what is not working? When I use the example from the test and enter it into the playground I already get the same result as the test.

```
import React from 'react';
export default React.memo(React.forwardRef(({ foo = 'bar' }, ref) => <div ref={ref}>{foo}</div>));
```

mfazekas:
> Thanks for contributing.
> 
> Could you show me an example of what is not working? When I use the example from the test and enter it into the playground I already get the same result as in the test.
> 
> ```
> import React from 'react';
> export default React.memo(React.forwardRef(({ foo = 'bar' }, ref) => <div ref={ref}>{foo}</div>));
> ```


What I see in playground, and in our case also is that this __does not work__ in playground: 
```jsx
import React, {memo,forwardRef} from 'react';


export default memo(forwardRef(({ foo = 'bar' }, ref) => <div ref={ref}>{foo}</div>));
```


This does work:
```jsx
import React, {memo,forwardRef} from 'react';


export default React.memo(forwardRef(({ foo = 'bar' }, ref) => <div ref={ref}>{foo}</div>));
```

So does this:
```jsx
import React, {memo,forwardRef} from 'react';


export default memo(({ foo = 'bar' }, ref) => <div ref={ref}>{foo}</div>);
```

I'll update the test code in the PR

PR #593, opened 2022-05-12
Handles ObjectTypeSpreadProperty in object type annotation.

### LINKED ISSUE #449: Did not recognize object of type "ObjectTypeSpreadProperty"
I am using the `react-styleguidist` package, which makes use of `react-docgen`. When I want to build my styleguide, I get the following error:

```
Warning: Cannot parse src/Sulu/Bundle/AdminBundle/Resources/js/components/Checkbox/Checkbox.js: Error: did not recognize object of type "ObjectTypeSpreadProperty"
```

I am using flow for static type checks in this project, and I assume the error is because of type definitions using the spread operator, in this specific case:

```javascript
type Props = {|
    ...SwitchProps,
    className?: string,
    onChange?: (checked: boolean, value?: string | number) => void,
    skin: 'dark' | 'light',
|};
```

Is there a reason this is not supported, or is this an error?

imdreamrunner:
This PR is attempting to fix #449

With the fix, react-docgen is able to support the following senario

```javascript
type OtherFruits = { orange: string };
type props = { apple: string, banana: string, ...OtherFruits };
```


danez (approved):
Awesome, thanks a lot

src/utils/__tests__/getFlowType-test.ts
```diff
@@ -1269,4 +1269,96 @@ describe('getFlowType', () => {
...
+    const typePath = statement(`
+      var x: {| apple: string, banana: string, ...MyType |} = 2;
+      import type { MyType } from 'MyType';
+    `)
+      .get('declarations', 0)
+      .get('id')
+      .get('typeAnnotation')
+      .get('typeAnnotation');
+
+    expect(getFlowType(typePath, null, mockImporter)).toEqual({
```
danez, line 1325: For the 2 new test can you please use `toMatchSnapshot()`.
I know the other tests don't use it, but using it keeps the tests smaller and readable.

danez (changes requested):

src/utils/__tests__/getFlowType-test.ts
imdreamrunner: Great suggestion!

danez: What was the reasoning to use `toMatchInlineSnapshot` in favor of `toMatchSnapshot`? Did the later one not work?

imdreamrunner: Oops sorry I mistakenly assumed that we wanted to put the snapshot together with the cases. Just updated to use `toMatchSnapshot` and found there're other test cases using `toMatchSnapshot` already 😄

danez (approved):

Issue #496, opened 2021-02-26
Cannot get comment under object in TypeScript


xiaoshuangLi:
The ```shape``` is ```Object```, the comment of ```shape.string``` and ```shape.number``` don't work.

```js
// react-docgen@5.3.1

const code = `
  import React from 'react';
  import PropTypes from 'prop-types';

  type Props = {
    /**
     * shape
     */
    shape: {
      /**
       * string
       */
      string: string;
      /**
       * number
       */
      number: number;
    };
  };

  class TSReactComponent extends Component<Props> {
    render() {
      return (
        <div {...this.props} />
      );
    }
  }

  export default TSReactComponent;
`;


const result = parse(code) || {};

/** 
 * result ==> json
 *
 * shape.string doesn't have description, should be { "description": "string" }
 * shape.number doesn't have description, should be { "description": "number" }
 */
const json = {
  "props": {
    "shape": {
      "required": true,
      "tsType": {
        "name": "signature",
        "type": "object",
        "raw": "{\n  /**\n   * string\n   */\n  string: string;\n  /**\n   * number\n   */\n  number: number;\n}",
        "signature": {
          "properties": [{
            "key": "string",
            "value": {
              "name": "string",
              "required": true
            }
          }, {
            "key": "number",
            "value": {
              "name": "number",
              "required": true
            }
          }]
        }
      },
      "description": "shape"
    }
  }
};
```

hosseini44444:
I've faced the same problem.
``` ts
interface MainOptions {
  /** If set to true video will start on load. */
  autoplay: boolean;
  /** If set to true video will repeat after end */
  repeat: boolean;
}
interface props {
  /** Main options of the player */
  options: Partial<MainOptions>;
}
```
But I don't get any description for `autoplay` and `repeat` params.
Any solutions?

PR #663, opened 2022-09-10
fix(ts): parse jsdoc comment for ts structs

mfazekas:
Fixes: #496

For something like:
```tsx
type Props = {
  ...
  center: {
    /** x coordinate */
    x : number,
    /** y coordinate */
    y: number;
  }
}
```

Fill out `description` fields with appropriate jdoc for `x` and `y`

### DIFF
--- a/packages/react-docgen/src/Documentation.ts
+++ b/packages/react-docgen/src/Documentation.ts
@@ -117,6 +117,7 @@ export interface ObjectSignatureType<T = FunctionSignatureType>
     properties: Array<{
       key: TypeDescriptor<T> | string;
       value: TypeDescriptor<T>;
+      description?: string;
     }>;
     constructor?: TypeDescriptor<T>;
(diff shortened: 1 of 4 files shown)


danez (changes requested):
Thank you for the PR and sorry for the late reply.

It looks good, only one thing:
Could you add the optional description to the type here: https://github.com/reactjs/react-docgen/blob/main/packages/react-docgen/src/Documentation.ts#L117-L120

Thanks

mfazekas:
> Thank you for the PR and sorry for the late reply.
> 
> It looks good, only one thing: Could you add the optional description to the type here: https://github.com/reactjs/react-docgen/blob/main/packages/react-docgen/src/Documentation.ts#L117-L120
> 
> Thanks

Thanks, fixed!

danez (approved):

PR #674, opened 2022-10-08
fix(ts): parse prop types if component inhereited from a HOC

mfazekas:
Supports doc generation for the following, pattern:

```
function MyHoc<T>(v: React.Component<T>) : React.Component<T> {
	return v;
}

interface Props {
  ...
}

class MyComponent extends MyHoc(React.Component<Props>) {
  render() {
    return <div className='s-c-z-example'/>;
  }
}
```

It's only a heuristic, and has the assumption that a function call in extends is a HOC and the first argument to Hoc is a React.Component like thing.

Note sure I this is something you'd support or if this implementation is ok or not.

### DIFF
--- a/packages/react-docgen/src/utils/getTypeFromReactComponent.ts
+++ b/packages/react-docgen/src/utils/getTypeFromReactComponent.ts
@@ -19,6 +19,8 @@ import type {
   TSTypeParameterInstantiation,
   TypeParameterDeclaration,
   TypeParameterInstantiation,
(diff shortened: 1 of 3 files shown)


danez:
Thanks for the PR. I'm not sure if this should be supported. I don't think it is a very common usecase and also it seems react class components are pretty much on their way out.

What is the actual usecase of doing this vs creating a separate actual component that gets the HOC applied?

PR #924, opened 2024-05-27
Add support for function components that return null

### LINKED ISSUE #926: No support for components that directly return `null`
If a component directly returns null like this:
```typescript
export const DataTableColumnDocs: FunctionComponent<
  IDataTableColumn<object>
> = () => {
  return null;
};
```

Then it won't find any interface to parse. The same behavior can be seen when a component directly returns a prop like so:
```typescript
export const AuthenticatedBoundary = ({
  children,
  fallback,
}: AuthenticatedBoundaryProps) => {
  const { isLoggedIn } = useAuthentication();

  if (!isLoggedIn) {
    return fallback;
  }

  return children;
};
```

rvetere:
Add support for function components that only return null:

```typescript
export const DataTableColumnDocs: FunctionComponent<
  IDataTableColumn<object>
> = () => {
  return null;
};
```

and add support for components that only return a prop (directly, without a fragment):
```typescript
export const AuthenticatedBoundary = ({
  children,
  fallback,
}: AuthenticatedBoundaryProps) => {
  const { isLoggedIn } = useAuthentication();

  if (!isLoggedIn) {
    return fallback;
  }

  return children;
};
```

Fixes #926


packages/react-docgen/src/utils/__tests__/findFunctionReturn-test.ts
```diff
@@ -152,6 +152,27 @@ describe('findFunctionReturn', () => {
...
+      `);
+
+      expect(findFunctionReturn(def, predicate)).toBeDefined();
+    });
+
+    test('handles null return', () => {
+      const def = parse.statement(`
+        function Foo (props) {
+          return null;
+        }
```
danez, line 169: What worries me is that literaly any function that returns null will be detected as react component.
Same for the other usecase.

PR #969, opened 2025-01-21
Expose more Typescript types

### LINKED ISSUE #864: Expose more TypeScript types
We are using react-docgen to parse TypeScript component files and display their props and types in a documentation app. Our issue is that typing our parser code is not possible since these types are not exported. Would it be possible to expose more types? Heres our code: https://github.com/instructure/instructure-ui/blob/56c7c38e3f360b7110a720a7268b770c916e1500/packages/__docs__/buildScripts/DataTypes.mts#L89C1-L89C1
 We'd like that these types are exported at least: 

```
interface MethodParameter 
interface MethodReturn 
interface PropDescriptor 
interface PropTypeDescriptor
type TypeDescriptor
interface SimpleType
interface LiteralType
interface ElementsType
interface FunctionArgumentType
interface FunctionSignatureType
interface TSFunctionSignatureType
interface ObjectSignatureType
interface BaseType
```
So basically the types used by the `Documentation` type. Thanks!

christon88:
This should fix #864

### DIFF
--- a/.changeset/curvy-roses-deny.md
+++ b/.changeset/curvy-roses-deny.md
@@ -1,0 +1,5 @@
+---
+"react-docgen": patch
+---
+
+Expose more TypeScript types
--- a/packages/react-docgen/src/main.ts
+++ b/packages/react-docgen/src/main.ts
@@ -1,25 +1,40 @@
+import type DocumentationBuilder from './Documentation.js';
+import type {
+  BaseType,
+  Documentation,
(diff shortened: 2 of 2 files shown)


christon88:
I realize now that my linter sorted the imports and exports alphabetically. If there is a different system for sorting imports and exports, or if you'd rather they remain as they were, I'd be happy to change it back!

PR #1103, opened 2026-07-14
Support Babel 8 dependency overrides

### LINKED ISSUE #1102: Support Babel 8 in babelParser (loadPartialConfigSync)
## Description

`react-docgen@8.0.3` cannot parse components when its `@babel/core` dependency is resolved to Babel 8.

`babelParser` calls `loadPartialConfig` synchronously. Starting with Babel 8, that API expects a callback; synchronous callers must use `loadPartialConfigSync`.

This surfaced through Storybook's `react-docgen` plugin when a workspace-wide override resolved Babel to 8.0.1. Storybook is currently tracking Babel 8 support and explicitly waiting for react-docgen compatibility: https://github.com/storybookjs/storybook/issues/35295

## Minimal reproduction

`package.json`:

```json
{
  "private": true,
  "type": "module",
  "packageManager": "pnpm@11.13.0",
  "dependencies": {
    "@babel/core": "8.0.1",
    "react-docgen": "8.0.3"
  }
}
```

`pnpm-workspace.yaml`:

```yaml
packages:
  - .

overrides:
  '@babel/core': 8.0.1
```

`repro.mjs`:

```js
import { parse } from 'react-docgen';

parse('export const Button = ({ label }) => <button>{label}</button>;');
```

Run:

```sh
pnpm install
node repro.mjs
```

## Actual result

```text
Error: Starting from Babel 8.0.0, the 'loadPartialConfig' function expects a callback. If you need to call it synchronously, please use 'loadPartialConfigSync'.
    at loadPartialConfig (.../@babel/core/lib/index-shared.js:2207:11)
    at buildPluginList (.../react-docgen/dist/babelParser.js:34:27)
```

## Expected result

Parsing succeeds with Babel 8, while retaining Babel 7 compatibility.

## Versions

- `react-docgen`: 8.0.3
- `@babel/core`: 8.0.1
- Node.js: 24.18.0
- pnpm: 11.13.0

The likely compatibility change is to use `loadPartialConfigSync` for this synchronous code path; that API is also available in Babel 7.

Stanzilla:
## Summary

Babel 8 dependency overrides fail because react-docgen uses removed parser options and path APIs, and reads TypeScript AST fields that changed in Babel 8.

Use the synchronous config API, preserve Babel 7 parser defaults, replace removed path checks, and support both AST shapes for generic props, inherited interfaces, function types, and mapped types.

Run all library and CLI tests in separate Babel 7 and Babel 8 CI jobs. The Babel 8 jobs replace core, traverse, and types, so CLI subprocesses use Babel 8 too. Keep all jobs and both test suites running when a test fails. Use Babel 8's minimum supported Node.js version on the Node 22 job.

The declared Babel 7 dependency ranges remain unchanged. Packages build against those declared types before the runtime compatibility tests. Contributor instructions include the Babel 8 reproduction commands.

## Validation

On macOS with Node.js 24.20.0:

- Babel 7: all 1,447 library tests and 59 CLI tests pass with coverage.
- Babel 8 (core 8.0.1, traverse/types 8.0.4): all 1,447 library tests and 59 CLI tests pass with coverage.
- Existing snapshots pass unchanged. The two type-parameter test fixtures now read the correct type-argument field for each Babel version.
- `pnpm build` and `pnpm lint` pass.

Closes #1102

### DIFF
--- a/.changeset/tidy-beans-smile.md
+++ b/.changeset/tidy-beans-smile.md
@@ -1,0 +1,5 @@
+---
+'react-docgen': patch
+---
+
+Allow parsing with Babel 8 when it is supplied via an override, while preserving Babel 7's default parser behavior.
--- a/packages/react-docgen/package.json
+++ b/packages/react-docgen/package.json
@@ -53,5 +53,8 @@
     "doctrine": "^3.0.0",
     "resolve": "^1.22.1",
     "strip-indent": "^4.0.0"
+  },
+  "devDependencies": {
+    "babel-core-8": "npm:@babel/core@8.0.1"
   }
 }
--- a/packages/react-docgen/src/__tests__/babel8-test.ts
+++ b/packages/react-docgen/src/__tests__/babel8-test.ts
@@ -1,0 +1,10 @@
+import { parse } from '../main.js';
+import { expect, test, vi } from 'vitest';
+
+vi.mock('@babel/core', () => import('babel-core-8'));
+
+test('parses components with Babel 8', () => {
+  const result = parse('export function Button() { return <button />; }');
+
+  expect(result).toHaveLength(1);
+});
--- a/packages/react-docgen/src/__tests__/babel8-test.ts
+++ b/packages/react-docgen/src/__tests__/babel8-test.ts
@@ -8,3 +8,31 @@ test('parses components with Babel 8', () => {
 
   expect(result).toHaveLength(1);
 });
+
+test('parses TypeScript function props with Babel 8', () => {
+  const result = parse(
+    `export type MenuProps = {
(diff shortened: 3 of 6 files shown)


danez:
Thanks, I looked into it a little and there seems to be still a bunch of stuff missing. The first thing we need to do is run all tests with babel8 in parallel to babel 7. To see what is breaking.

Stanzilla:
> Thanks, I looked into it a little and there seems to be still a bunch of stuff missing. The first thing we need to do is run all tests with babel8 in parallel to babel 7. To see what is breaking.

Sorry for the delay, I just picked this up and am willing to implement it end-to-end, please let me know if you have any preferences or if "fixing what broke" is good enough!

PR #1139, opened 2026-09-20
fix: detect useCallback methods in imperative handles

### LINKED ISSUE #856: useCallback + useImperativeHandle  in methods
Methods in the following component is not recognised. It works if I remove the useCallback.

```jsx
import React, {
  forwardRef,
  memo,
  useCallback,
  useImperativeHandle,
  useMemo,
  useRef,
} from 'react';

export const MyComponent =
memo(forwardRef((_, ref) => {

  const _myMethod = useCallback((argument:string) : number => {});
  //const _myMethod = (argument:string) : number => {};
  
  useImperativeHandle(
    ref,
    () => ({
        /** myMethod description */
        _myMethod,
    }),
    [],
  );

  return <div />;
}));
```

askalf:
Fixes #856

## Summary
- Resolve the function wrapped by React `useCallback` when documenting component methods.
- Use that shared resolution for both method admission and signature extraction.
- Add coverage across imperative handles, object/statics/class/assignment surfaces and negative controls.

## Decisions
Resolution intentionally unwraps one React `useCallback` layer only; recursively unwrapping or treating other React builtins as methods would broaden behavior beyond this defect. The package suite and formatting checks passed.

AI assistance: this bug was found and the fix and tests were drafted with AI tooling in my workflow; the tests and checks above were executed as pasted. I'm responsible for the change and will handle review feedback.


packages/react-docgen/src/handlers/__tests__/componentMethodsHandler-useCallback-test.ts
```diff
@@ -0,0 +1,321 @@
...
+      imports:
+        "import { useCallback as useCb, useImperativeHandle } from 'react';",
+      setup: '',
+      value: 'useCb((argument: string): number => 1, [])',
+      imperativeHandle: 'useImperativeHandle',
+    },
+  ])(
+    'extracts a method wrapped with $name',
+    ({ imports, setup, value, imperativeHandle }) => {
+      const definition = parse.statementLast<FunctionDeclaration>(`
```
danez: Could we add a regression test that runs the exact shape from #856 through the normal parser/resolver: `memo(forwardRef(...))`, a top-level `useCallback`, and shorthand exposure through `useImperativeHandle`? These tests invoke the handler directly with a plain function, so they do not pin the wrapper-resolution path from the reported bug.

_AI disclosure: This comment was added by ChatGPT 5.6 Sol._

packages/react-docgen/src/handlers/componentMethodsHandler.ts
```diff
@@ -40,9 +42,8 @@ function isMethod(path: NodePath): path is MethodNodePath {
     !isProbablyMethod &&
     (path.isClassProperty() || path.isObjectProperty())
   ) {
-    const value = resolveToValue(path.get('value') as NodePath);
-
-    isProbablyMethod = value.isFunction();
+    isProbablyMethod =
+      resolveToMethodFunction(path.get('value') as NodePath) !== null;
```
danez: This changes generic object and class-property method detection, although #856 only asks about methods exposed through `useImperativeHandle`. It also adds support for `useCallback` in class fields/statics, where Hooks are not valid React usage. Could we keep the unwrapping local to the imperative-handle path unless there is a separate use case for broadening every method surface?

_AI disclosure: This comment was added by ChatGPT 5.6 Sol._

packages/react-docgen/src/utils/__tests__/getMethodDocumentation-useCallback-test.ts
```diff
@@ -0,0 +1,46 @@
...
+import getMethodDocumentation from '../getMethodDocumentation.js';
+import { describe, expect, test } from 'vitest';
+
+describe('getMethodDocumentation useCallback', () => {
+  test('documents the function a useCallback assignment wraps', () => {
+    const method = parse
+      .statementLast<ExpressionStatement>(
+        `import { useCallback } from 'react';
+         const Foo = () => {}
+         Foo.foo = useCallback((bar: number): number => bar, [])
```
danez: This test suggests that `Component.foo = useCallback(...)` is supported, but `findAssignedMethods` still filters assignments with `resolveToValue(right).isFunction()`, so the public handler never passes this assignment to `getMethodDocumentation`. Please either update the handler consistently and test it through `componentMethodsHandler`, or remove this out-of-scope assignment case.

_AI disclosure: This comment was added by ChatGPT 5.6 Sol._

danez (changes requested):
Thanks for the PR.

The core `useImperativeHandle` + `useCallback` support makes sense, but I think the tests and implementation should be tightened before merging. In particular, please cover the exact issue reproducer and either narrow or consistently implement the additional method surfaces.

packages/react-docgen/src/handlers/__tests__/componentMethodsHandler-useCallback-test.ts
askalf: Added in c97359c: `src/__tests__/imperativeHandleCallbacks-test.ts` parses the component from #856 verbatim (memo(forwardRef(...)), a top-level useCallback, shorthand exposure through useImperativeHandle) through the public parse(), alongside a plain-function memo(forwardRef) case. Both fail on main without the fix.

packages/react-docgen/src/handlers/componentMethodsHandler.ts
askalf: Narrowed in f5d7939. The useCallback unwrapping now lives only on the imperative-handle path: getMethodFunctionExpression keeps its old plain-function resolution and consults useCallback only when the caller passes isImperativeHandle, which the handler sets from the useImperativeHandle visitor alone. Class fields, statics and object methods are back to base behaviour, and there are tests pinning that they stay there.

packages/react-docgen/src/utils/__tests__/getMethodDocumentation-useCallback-test.ts
askalf: Removed in e9a51c5. The Component.foo = useCallback(...) case was out of scope and the handler never reached it; the test file is gone rather than the handler widened.

askalf:
Thanks for the review. Pushed a9605a4..1d45d98, one commit per point: the exact #856 shape now runs through the parser (c97359c), the unwrapping is confined to the imperative-handle path (f5d7939), and the assignment case is dropped (e9a51c5); 1d45d98 pins the lifecycle filter on an unwrapped callback. Package suite: 1467 tests pass, prettier, eslint and tsc clean.

